PR #6, opened 2016-06-10
added Translator for Wink Binary Switches

rlingineni:


org.OpenT2T.Sample.SuperPopular.Outlet/Wink Binary Switch/js/thingTranslator.js
```diff
@@ -0,0 +1,113 @@
...
+        console.log('disconnect called.');
+        logDeviceState(this.device);
+    }
+};
+
+// globals for JxCore host
+global.initDevice = module.exports.initDevice;
+global.turnOn = module.exports.turnOn;
+global.turnOff = module.exports.turnOff;
+global.setBrightness = module.exports.setBrightness;
```
tjaffri: there is no set brightness...

org.OpenT2T.Sample.SuperPopular.Outlet/Wink Binary Switch/js/thingTranslator.js
```diff
@@ -0,0 +1,113 @@
...
+    if (typeof (device) !== 'undefined') {
+        console.log('  device.name          : ' + device.name);
+        console.log('  device.props         : ' + device.props);
+    } else {
+        console.log('device is undefined');
+    }
+};
+
+// Helper method to send power state commands to Wink
+function sendPowerStateCommandToDevice(powerState) {
```
tjaffri: This looks like it is pretty common across a lot of Wink-type translators... no change for now, but I will set up time with you to pair-program a "helper" library that we can use to refactor these API calls into a single node module. We could fork this https://github.com/TheThingSystem/node-winkapi or start fresh if we need a different type of interface.

rlingineni: Agreed. I was gonna ask the same, seems like all Wink devices could easily be mapped because they share a common set of characteristics. A helper library would be useful.

PR #7, opened 2016-06-11
Hue translators with promises, TypeScript

jasongin:
Adding 2 variations of the Hue lamp translator. The "js-promise" variation
returns promises from the methods for better async benavior. The "ts"
variation is also async and uses TypeScript for type-checking and async/await
with promises. (Async translator methods require bridge changes that are
not yet merged.)


org.OpenT2T.Sample.SuperPopular.Lamp/Philips Hue Light Bulb/ts/test.js
```diff
@@ -0,0 +1,41 @@
+var translator = require('./thingTranslator.js');
+var q = require('q');
```
tjaffri: I was hoping the test file could also be TypeScript, so we could show direct consumers of this translator how to call it from typescript (e.g. if somebody is writing a bot or website in ts they could refer to the test.ts file to see how to call it). Any chance you could update this file to ts?

org.OpenT2T.Sample.SuperPopular.Lamp/Philips Hue Light Bulb/ts/thingTranslator.ts
```diff
@@ -0,0 +1,108 @@
+
+declare function require(module: string): any;
```
tjaffri: shouldn't this be an import, not a require? I am not a typescript expert, but I thought that's how you do it in typescript? See https://basarat.gitbooks.io/typescript/content/docs/quick/nodejs.html

jasongin: Import doesn't work when there are no type definitions. I could have created (partial) type definitions for node-hue-api, but I wanted to demonstrate how that isn't necessary.

jasongin:
I added a commit which converts test.js to TypeScript, and automatically sets up typings for a few things: require, q, optimist.

PR #8, opened 2016-06-12
Added Thermostat Schema

rlingineni:


org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
```diff
@@ -0,0 +1,33 @@
...
+     <!-- Turn off the HVAC -->
+    <method name="turnOff"/>
+    
+    <!-- Get the current temperature -->
+    <method name="getCurrentTemperature">
+      <arg name="temp" type="d" direction="out" />
+    </method>
+   
+   <!-- set the current temperature -->
+    <method name="setCurrentTemperature">
```
jasongin: This should be named something different, since setting a temperature on the thermostat does not set the "current" temperature, that is the ambient temperature measured by the thermostat. I think the term "setpoint" is most often used for setting the desired temperature on a thermostat. And the schema should support retrieving the setpoint separately from the current temperature.

 Also, most thermostats support separate setpoints for heating and cooling.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
```diff
@@ -0,0 +1,33 @@
...
+      <arg name="temp" type="d" direction="in" />
+    </method>
+   
+   <!-- set the mode (e.g. Heating, Cooling, Auto)-->
+    <method name="setMode">
+      <arg name="mode" type="d" direction="in" />
```
jasongin: I would expect the mode arg type to be "s" (string). And it should be possible to get the current mode value too.

Also, I think there will need to be a way to get the set of available mode values, because different thermostats support different modes.

jasongin:
There are some unresolved merge conflicts in "Wink Binary Switch/js/test.js". Were those other files even intended to be included in the PR?

rlingineni:
Woops. Nope, Only the readme and the schema.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
rlingineni: Oh yeah, type 's' for sure.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
rlingineni: Hmm... I agree with removing the "Current", but I think "SetTemperature" is necessary. We could probably break it up as "SetHeatingTemperature" and "SetCoolingTemperature".

jasongin: I think "get/setHeatSetpoint", "get/setCoolSetpoint" would be clearer because "setpoint" is the standard term for what this is doing. Reference the first sentence on the [Wikipedia Thermostat page](https://en.wikipedia.org/wiki/Thermostat).

rlingineni: Hmm. I guess you make a point. I didn't realize this was an industry standard. I agree with setpoint then.

rlingineni: Done. Modified the Schema, let me know what you think.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
```diff
@@ -0,0 +1,48 @@
...
+      <arg name="temp" type="d" direction="in" />
+    </method>
+    
+    <!-- set the desired heating temperature (a.k.a. min setpoint) -->
+    <method name="setHeatingSetPoint">
+      <arg name="temp" type="d" direction="in" />
+    </method>
+    
+   <!-- set the mode (e.g. Heating, Cooling, Auto)-->
+    <method name="setMode">
```
jasongin: I think an application would need a method to get the mode, and a way to list all the modes supported by a particular thermostat. But those things can be added on later I guess.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
```diff
@@ -0,0 +1,48 @@
...
+  <interface name="org.OpenT2T.Sample.SuperPopular.Thermostat">
+  
+    <!-- Turn on the HVAC -->
+    <method name="turnOn"/>
+    
+     <!-- Turn off the HVAC -->
+    <method name="turnOff"/>
+    
+    <!-- Get the current ambient temperature -->
+    <method name="getRoomTemperature">
```
jasongin: I actually liked "getCurrentTemperature" (what you had before) better. I think "getRoomTemperature" could be a bit confusing in some cases. For example my thermostat (an ecobee) can report separate temperature readings for multiple rooms and average (a subset of) them together to get the "current" temperature.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
rlingineni: Yes you're right, I'll add those methods as well.

org.OpenT2T.Sample.SuperPopular.Thermostat/org.OpenT2T.Sample.SuperPopular.Thermostat.xml
```diff
@@ -0,0 +1,48 @@
...
+      <arg name="temp" type="d" direction="out" />
+    </method>
+
+     <!-- get the current cooling temperature (a.k.a. max setpoint) -->
+    <method name="getCoolingSetPoint">
```
jasongin: I'm nitpicking, but the 'P' should not be capitalized because "setpoint" as used here is one word.

tjaffri:
I opened this issue on the bridge to support collections eventually: https://github.com/tjaffri/msiot-samples/issues/5

That won't happen immediately, so till then I propose merging this PR. @jasongin, what do you think? If you're good with it I will merge.

jasongin:
I think we should go ahead and add a getMode method now.

As @tjaffri suggests, a getAvailableModes method can wait until we have support for arrays.

rlingineni:
I just added a getMode, we should be good to go with this schema :)

jasongin:
👍

rlingineni:
Just added the Wink-Helper and did the Wink Thermostat using the Wink-Helper. You will need to modify the local path in the thermostat translator to point to the helper. You can now write a translator for wink in roughly 2 minutes  :)

Also, instead of `getAvailableModes`, I'm thinking we could just have three hard set methods like ' `setCooling` , `setHeatingMode`,  `setAuto` .. I think it's fairly standard

jasongin:
I don't think I like having separate methods for setting the different modes: some thermostats support an Auto mode and some don't, some don't support cooling because there's no AC installed, etc. And my previous thermostat also defined additional modes called "Program Auto/Program Heat/Program Cool".

I actually implemented arrays in the bridge last night; I'll submit a PR for that soon. (Trying to get dictionaries working also.) So `getAvailableModes` will be unblocked. I think you can go ahead and write the schema entry and code for it now anyway, because it is testable via node command-line. BTW the schema type syntax for "array of strings" is not "a*s" as I wrote before but just "as".

jasongin:
Oh, and as for the Wink helper and translator code, can you move those to a separate PR? I will have some review comments on that code, but I'll wait until it's separated. I think we'd like to go ahead and get the schema in sooner so that others can work on other thermostat translators.

rlingineni:
I just added the 'getAvailableModes' method, we should be good to go there. As per the Wink-Helper, I'm not really familiar with git, is it possible I can create a separate PR from my commit?

tjaffri:
Hey Ravi... could you please do one PR with only the schema (just the XML) and a separate PR with the translator and the Wink Helper? I agree with @jasongin that we should review those separately.

I think the schema is ready to merge... we should review the translator and Wink Helper a bit more in a separate PR.

Thanks!

rlingineni:
Yes, for sure. I'm just figuring out how to do that. Jason is helping me out with splitting up the commits. 

Thanks,
Raviteja.L

jasongin:
👍 
I helped @rlingineni separate the wink code into another PR. This one looks good to go now.

rlingineni:
Thanks  @jasongin

PR #16, opened 2016-06-15
Added wink helper and thermostat

rlingineni:


org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,169 @@
+'use strict';
+
+//var request = require('request');
```
tjaffri: remove?

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,169 @@
...
+var https = require('https');
+var q = require('q');
+
+//this is our base, we each call has its own set of options
```
tjaffri: typo?

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,169 @@
...
+    return deferred.promise;
+    
+   },
+   
+   getLastReading: function(apiField,callback)
+   {
+        var deferred = q.defer();   // q will help us with returning a promise
+    
+    
+    //the headers to make our call
```
tjaffri: Something seems to be up with formatting on this file... can you fix it so things are aligned correctly?

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "opent2t-translator-wink-thermostat",
+    "version": "1.0.0",
+    "description": "",
+    "dependencies": {},
```
jasongin: This package needs to list opent2t-translator-helper-wink in its dependencies.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
```diff
@@ -0,0 +1,146 @@
+'use strict';
+
+var https = require('https');
+var wh = require("org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js"); //change this to your local
```
jasongin: This should require the helper via an npm dependency: `require('opent2t-translator-helper-wink')`
Note you can `npm install` a package from a local path even before it is published to npm.

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,169 @@
...
+{
+    //get the initial parameters to build our target endpoint
+   initWinkApi: function(apiEndpoint, deviceID, token)
+   {
+       options.path = '/' + apiEndpoint + '/' + deviceID;
+       accessToken = token; //the specific token
+        
+   },
+ 
+   sendDesiredStateCommand: function(apiField,value,callback)
```
jasongin: There's no need for all of these methods to take a callback. If a caller wants to do something after the async operation they can call .then() on the returned promise.

rlingineni: hmm ok. so should I just remove the callback as a parameter? 

For example in the test.js right now it works kind of like:
```
wh.getLastReading('someparam',function (error, result) //last reading fields are read-only
{
    console.log(result); 
});
```
I'm not familiar with promises, so how would it work in that case using the ".then()"?

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,169 @@
...
+        {
+             deferred.resolve('ended call');
+        });
+        res.on('error', (e) => {
+            deferred.reject(e.message);
+        }); });
+
+    req.on('error', (e) => {
+        console.log('problem with request:');
+        deferred.reject(e.message);
```
jasongin: Pass the Error object to reject(), not just the message.

PR #18, opened 2016-06-16
Added Wink Helper and  Wink Thermostat

rlingineni:
Wink Thermostat uses the Wink helper


org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,167 @@
...
+
+       var body = ' '; //holds the chunks of data
+        res.on('data', (chunk) => 
+        {
+          body += chunk; 
+        });
+        res.on('end', () => 
+        {
+            //parse the JSON response and look for the apiField we want in the JSON body
+             var results = JSON.parse(body.toString());
```
jasongin: Wrap the JSON.parse call in a try/catch block, then reject the promise if parsing failed.

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,167 @@
...
+        'Accept': 'application/json',
+        'Content-length':  JSON.stringify(options.postData).length
+    }
+    
+    options.method = 'PUT'; //the type of request to make
+   
+    //our HTTPS call 
+    var req = https.request(options, (res) => {
+        res.setEncoding('utf8');
+        res.on('data', (chunk) =>
```
jasongin: This callback doesn't do anything, so can it be deleted?

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
rlingineni: Ah yes, good idea

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
rlingineni: Callback? Which one?

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,167 @@
+'use strict';
+
+var https = require('https');
+var q = require('q');
+
+//this is our base, we refactor these options in each method
+ var options =
```
jasongin: The options object should not be global. It will cause problems when concurrent calls are trying to use the same object. Instead, create a new options object within the scope of each call. You can still define these strings separately as constants at the top if you want.

rlingineni: Ok. Will change that to create a new object

jasongin: Or if you want to have a shared options object that you use to initialize the options for each call, use something like [node-clone](https://github.com/pvorb/node-clone) to make a copy before modifying it to be specific to each call.

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,167 @@
...
+       var body = ' '; //holds the chunks of data
+        res.on('data', (chunk) => 
+        {
+          body += chunk; 
+        });
+        res.on('end', () => 
+        {
+            //parse the JSON response and look for the apiField we want in the JSON body
+             var results = JSON.parse(body.toString());
+             var valueToGet = results.data.desired_state[apiField];
```
jasongin: Also include this line in the try/catch in case the results did not have the expected structure.

org.OpenT2T.Sample.SuperPopular.Helpers/js/test.js
```diff
@@ -0,0 +1,16 @@
+
+
+var wh = require('./index');
+
+wh.initWinkApi("endpointHere","deviceID","accesstoken");
+
+ wh.sendDesiredStateCommand('apiField',value).then(result => {
+   console.log(result);}).catch(error => {console.log(error.message);}); //for logging
```
jasongin: For better readability I'd suggest formatting with some more newlines:
```
wh.sendDesiredStateCommand('apiField', value).then(result => {
    ...
}).catch (error => {
    ...
});

File: org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js [hunk 6]
```diff
@@ -0,0 +1,138 @@
...
+        console.log('device is undefined');
+    }
+};
+
+var deviceId, accessToken;
+
+// module exports, implementing the schema
+module.exports = {
+
+    device: null,
```
jasongin: Delete this field. (I realize you copied this from somewhere else.) All the `this.device` references below don't even reference it anyway when running in a JXCore environment, because JXCore doesn't set `this` when calling the methods, so it defaults to the global object... So it happens to work, but it doesn't actually use this field.

I'm not sure you even need any references to `this.device` anyway, because `logDeviceState` doesn't seem to serve a useful purpose.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
```diff
@@ -0,0 +1,138 @@
...
+        wh.initWinkApi("thermostats",deviceId,accessToken);
+        console.log('Javascript and Wink Helper initialized.');
+        logDeviceState(this.device);
+    },
+
+    turnOn: function() {
+        console.log('turnOn called.');
+
+        wh.sendDesiredStateCommand('powered',true).then(result => {
+        console.log(result);}).catch(error => {console.log(error.message);});
```
jasongin: This needs to return the promise, and also after logging return the result / re-throw the error. Since this pattern is repeated many times, how about a helper function:
````
function logPromise(p) {
    return p.then(result => {
        console.log(result);
        return result;
    }).catch(error => {
        console.log(error.message);
        throw error;
    });
}
```

Then use it like this:
```
return logPromise(wh.sendDesiredStateCommand('powered', true));

File: org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/test.js [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+
+// Go through a sequence of test operations for the translator
+q.delay(1000)
+ .then(() => translator.turnOn())
+ .then(() => q.delay(2000))
+ .then(() => translator.setMode("heat_only"))
+ .then(() => q.delay(2000))
+ .then(() => console.log(translator.getCurrentTemperature()))
+ .then(() => q.delay(2000))
+ .then(() => translator.setCoolingSetpoint(23))
```
jasongin: My first thought was, "wow that's cold." Then I realized you're thinking Celsius.. and now I'm wondering if the thermostat schema needs to have something in it about units (Celsius vs Fahrenheit).

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
jasongin: res.on('data', (chunk) => {   });

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
```diff
@@ -0,0 +1,138 @@
...
+                if (typeof (props.access_token) !== 'undefined') {
+                    accessToken = props.access_token;
+                } else {
+                    console.log('props.access_token is undefined.');
+                }
+
+                if (typeof (props.id) !== 'undefined') {
+                    deviceId = props.id;
+                } else {
+                    console.log('props.id is undefined.');
```
jasongin: For all these error cases, you should `throw new Error("...")` instead of just logging.

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
rlingineni: I would keep it there, it's consistent with all of the other HTTPS requests, and there maybe some calls in which we want to log from it.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
rlingineni: Good idea. 😃 I should've thought of this before I repeated my code like 10 times

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/test.js
rlingineni: We could do this effectively with two more functions in the schema. One would be a Boolean `isCelsius` and then if that is `false` then another method in the schema is `convertToCelsius` and all of the temperature methods in the translator would call that and we would make it work. Since `convertToCelsius` isn't a translator specific function, I'm thinking if there is a way to just kind of bundle that method as a standard without re-implementing it in every translator.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
rlingineni: I think this is something that all of the translators might need then

jasongin: In my PR I wrote a simple argument validation function here: https://github.com/openT2T/translators/pull/7/files#diff-27473d578f19f2381f47a9cf0bcafa50R33

org.OpenT2T.Sample.SuperPopular.Helpers/js/index.js
```diff
@@ -0,0 +1,167 @@
+'use strict';
```
tjaffri: I was hoping the folder name could be only "helpers" and inside that folder, put a sub-folder called opent2t-translator-helper-wink (i.e. the name of the node module). Namespacing this folder name like the other translators is likely to confuse people about what this is.

rlingineni: Sure, that works

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
tjaffri: Looks like this is an opportunity to start a NEW helper, i.e. helpers\opent2t-translator-helper. This should contain helper methods like logPromise and the argument validation function... things we need to share across all translators. @rlingineni, can you add that as part of this refactoring? @jasongin what do you think?

rlingineni: Yeah I could look into adding that under this branch

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/test.js
jasongin: For now, I think the schema should just document that the units are Celsius. The app and/or translator can do the conversion as needed. (Make sure it is a decimal type so there's no loss of precision.)

Longer term, there probably needs to be a discussion about how to handle units in general.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
jasongin: Yeah I wouldn't block this PR on that. We can always refactor out common code later. I agree eventually we'll probably want to have a helper module used by all translators. But it's probably a bit early for that now.

For validation, ideally we could just use some popular node argument-validation module... but surprisingly I couldn't find one.

For logging, I think we haven't figured out yet what should be the best practices for translators. The logging is really only beneficial for testing, not in a real application environment. For testing I think the logging could easily be much more verbose -- for example it could show the entire HTTP request/response headers and body, so when something goes wrong you can see all the details. But it should be disabled in an application environment.

rlingineni:
Take a look, tested it, and it works. Implemented what we discussed

Helpers/OpenT2T-Wink-Helper/js/index.js
```diff
@@ -0,0 +1,197 @@
...
+       console.log("Api path is " + ApiPath)
+       accessToken = token; //the specific token
+        
+   },
+ 
+   sendDesiredStateCommand: function(apiField,value) 
+   {
+       
+    var options = 
+        {
```
jasongin: Please make the indentation consistent throughout this file (4 spaces). Currently various blocks are using 1, 2, 3, or 4 spaces.

Helpers/OpenT2T-Wink-Helper/js/package.json
```diff
@@ -0,0 +1,15 @@
...
+    "version": "1.0.0",
+    "description": "",
+    "dependencies": {
+        "async": "~1.5.2",
+        "request":"2.72.0",
+        "q":"1.4.1"
+    },
+    "devDependencies": {
+         "optimist": "0.6.1" 
+    },
```
jasongin: You can remove this, it's not used anywhere in this package.

Helpers/OpenT2T-Wink-Helper/js/package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "opent2t-translator-helper-wink",
+    "version": "1.0.0",
+    "description": "",
+    "dependencies": {
+        "async": "~1.5.2",
+        "request":"2.72.0",
```
jasongin: Remove, it's not used.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "opent2t-translator-wink-thermostat",
+    "version": "1.0.0",
+    "description": "",
+    "dependencies": {
+        "opent2t-translator-helper-wink": "1.0.0"
+    },
+    "devDependencies": {
+        "async": "~1.5.2",
+        "optimist": "0.6.1",
```
jasongin: Remove, it's not used.

Helpers/OpenT2T-Wink-Helper/js/index.js
```diff
@@ -0,0 +1,197 @@
+'use strict';
+
+var https = require('https');
+var q = require('q');
+
+//this is our base, we refactor these options in each method
+
+
+var  accessToken,ApiPath;  //we will need this when creating the headers
```
jasongin: JavaScript naming convention is to only capitalize constructors/class names. So make this lowercase.

Helpers/OpenT2T-Wink-Helper/js/index.js
```diff
@@ -0,0 +1,197 @@
...
+        
+   },
+ 
+   sendDesiredStateCommand: function(apiField,value) 
+   {
+       
+    var options = 
+        {
+        protocol: 'https:',
+        host: 'api.wink.com',
```
jasongin: Define this string in one place the top.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/js/thingTranslator.js
```diff
@@ -0,0 +1,138 @@
...
+
+    getCurrentTemperature: function() {
+        console.log('getting current temp');
+
+        return logPromise(wh.getLastReading('temperature'));
+    },
+    
+    getCoolingSetpoint: function()
+    {
+        console.log('getting cooling point');
```
jasongin: JS open braces go at the end of the previous line.

Helpers/OpenT2T-Wink-Helper/js/index.js
```diff
@@ -0,0 +1,197 @@
...
+var q = require('q');
+
+//this is our base, we refactor these options in each method
+
+
+var  accessToken,ApiPath;  //we will need this when creating the headers
+      
+module.exports = 
+{
+    //get the initial parameters to build our target endpoint
```
jasongin: Open braces go at the end of the previous line.

Helpers/OpenT2T-Wink-Helper/js/index.js
rlingineni: which string?

tjaffri:
@rlingineni what's your ETA to incorporate @jasongin feedback? I'll merge once this last batch of feedback is incorporated. Thanks!

rlingineni:
@tjaffri I just added those changes. Should be good to go 👍

jasongin:
:shipit:

tjaffri:
Hah... loved the :shipit:... @sanjaiganesh I am merging this now.

jasongin:
@tjaffri CodeFlow put the :shipit: there when I signed off, not me :) I don't know why, I've only seen it use :+1: before.

PR #23, opened 2016-06-23
Wink Doorbell

rlingineni:
Works with the Ring Doorbell. Dependent on #18. Should merge that first.


rlingineni:
Dependent on #18. Should merge that first.

tjaffri:
@sanjaiganesh @jasongin any feedback on this before I merge? Would like a :shipit: squirrel (I get it now!) please.

org.OpenT2T.Sample.SuperPopular.Doorbell/org.OpenT2T.Sample.SuperPopular.Doorbell.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://standards.freedesktop.org/dbus/introspect-1.0.dtd"[]>
+<node>
+  <interface name="org.OpenT2T.Sample.SuperPopular.Doorbell">
+    
+   <!-- is the doorbell pressed? -->
+    <method name="isButtonPressed">
```
jasongin: Shouldn't this be a signal? Apps aren't going to constantly poll to check if a doorbell is pressed.

I know we don't yet have any example translators using signals. (I'm not even sure how well they work.)

jasongin:
@tjaffri I just have the one concern about proper use of signals. Obviously the entire purpose of a doorbell is to be a signal. But it's fine with me if we want to defer that issue until later.

tjaffri:
I'll open an issue to track the signals thing

PR #29, opened 2016-07-08
Initial checkin for the vera dimmable light translator.

arjun-msft:


tjaffri:
@jasongin you want to check this out and merge, please? I thought you said you have this setup at home so that might work out nicely. Thanks.

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+        res.setEncoding('utf8');
+
+        res.on('data', (chunk) => {
+        });
+
+        res.on('end', () => {
+            if (res.statusCode != 200)
+            {
+                console.log(res);
+                deferred.reject(res);
```
sanjaiganesh: I feel we need to come up with common error object that OpenT2T translators will throw. On error, we do deferred.reject(e). And In Nest  helper I do deferred.reject(new Error("Error Code:" + res.statusCode)), if statusCode!=200.

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+    req.on('error', (e) => {
+        console.log('problem with request:' + e.message);
+        deferred.reject(e);
+    });
+
+    req.on('response', (res) => {
+        res.setEncoding('utf8');
+
+        res.on('data', (chunk) => {
+        });
```
sanjaiganesh: The current pattern is, we usually capture the chunk here and then use parsed data in deferred.resolve() on' end'. you may take a look at nest helper (translators\helper\..).

sanjaiganesh:
:clock1:

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
sanjaiganesh: You may also want to refactor into a separate helper, like we have for Nest\Wink under translators\helpers

---
In reply to: [70480431](https://github.com/openT2T/translators/pull/29#discussion_r70480431) [](ancestors = 70480431)

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+    relaySessionToken : null,
+
+    // serial number of the vera hub
+    pkDevice : null,
+
+    // data structure which represents the device targetted by this translator
+    device: null,
+
+    initDevice: function(dev) {
+        var deferred = q.defer();
```
jasongin: You're not doing anything async in this function, so there's no need to use a promise here. Just throw Errors for any invalid arguments.

(This argument validation function)[[https://github.com/openT2T/translators/blob/master/org.OpenT2T.Sample.SuperPopular.Lamp/Philips%20Hue%20Light%20Bulb/js-promise/thingTranslator.js#L33](https://github.com/openT2T/translators/blob/master/org.OpenT2T.Sample.SuperPopular.Lamp/Philips%20Hue%20Light%20Bulb/js-promise/thingTranslator.js#L33)] may be helpful.

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+
+    // sets the dimmable bulb to the desired brightness, valid values: integer 0-100
+    setBrightness: function(brightness) {
+        console.log('[' + this.deviceId + '] setBrightness called with value: ' + brightness);
+        return sendPowerStateCommandToDevice(brightness, this.deviceId, this.relayServer, this.relaySessionToken, this.pkDevice);
+    },
+
+    disconnect: function() {
+        console.log('disconnect called.');
+        var deferred = q.defer();
```
jasongin: Again, no need for a promise here. (For all the methods, the caller will detect whether or not a promise was returned.)

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+
+// Helper method to send power state commands to Vera
+function sendPowerStateCommandToDevice(loadTarget, deviceId, relayServer, relaySessionToken, pkDevice) {
+    var deferred = q.defer();
+
+    // set up the url which needs:
+    //   the serial number of the hub
+    //   deviceid of the light bulb
+    //   target brightness
+    var path = '/relay/relay/relay/device/' + pkDevice + '/port_3480/data_request?id=action&output_format=json&DeviceNum=' + deviceId + '&serviceId=urn:upnp-org:serviceId:Dimming1&action=SetLoadLevelTarget&newLoadlevelTarget=' + loadTarget;
```
jasongin: Please wrap this long line. The GitHub web UI cuts off long lines.

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
jasongin: The argument passed to `reject()` should always be an `Error` instance.

I don't think it's likely we'll ever handle different kinds of errors differently -- that practice is not common in JavaScript anyway. The best you can do is provide a reasonable error message in the Error, since that will typically be propagated back in the exception to the caller using other languages.

---
In reply to: [70480151](https://github.com/openT2T/translators/pull/29#discussion_r70480151) [](ancestors = 70480151)

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
jasongin: For on/off commands (not dimming), would it be better to use the SwitchPower1 service? That way if you turn a light off and on again then (I think) it will come on to the previous brightness rather than always 100%.

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
```diff
@@ -0,0 +1,174 @@
...
+
+        console.log('Javascript initialized.');
+        logDeviceState(this.device);
+
+        return deferred.promise;
+    },
+
+    turnOn: function() {
+        console.log('[' + this.deviceId + '] turnOn called.');
+        return sendPowerStateCommandToDevice(100, this.deviceId, this.relayServer, this.relaySessionToken, this.pkDevice);
```
jasongin: Unfortunately, these `this.` references won't work when running in JXCore, at least with the current implementation. That's because the JXCore API for invoking a JS function doesn't have a way to specify the implicit `this` parameter. But probably we should fix that rather than having all the translators implement a workaround.

jasongin: Well, they might accidentally work, but only because `this` will refer to the `global` object. So those declared properties on the module above aren't actually used.

---
In reply to: [70485538](https://github.com/openT2T/translators/pull/29#discussion_r70485538) [](ancestors = 70485538)

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/test.js
```diff
@@ -0,0 +1,29 @@
...
+function Device(deviceId, relaySessionToken, serverRelay, pkDevice) {
+    this.props = ' { "id": "' + deviceId + '", "relay_session_token": "' + relaySessionToken + '", "relay_server": "' + serverRelay + '", "pk_device": "' + pkDevice + '" }';
+    this.name = "Vera Light Bulb (Test)";
+}
+
+var device = new Device(argv.i, argv.a, argv.r, argv.p);
+
+// initialize the translator for testing purposes (this is normally called by the runtime)
+translator.initDevice(device)
+    .then(function() { translator.turnOn(); })
```
jasongin: Return the promise from each of these functions for proper async chaining:
```
.then(function () { return translator.turnOn(); })

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
arjun-msft: I tried that originally and that did not work for me, can you double check your setup, maybe it depends hub to hub? or maybe I'm missing some other value.  

I'll change it to SwitchPower1 since that is more correct.

arjun-msft: Tried multiple things but the last value is not remember.  The physical switch remembers but Vera does not.

jasongin: OK, I wasn't sure about that. (I don't normally use my Vera with any lighting, mostly just shades.)

---
In reply to: [70526340](https://github.com/openT2T/translators/pull/29#discussion_r70526340) [](ancestors = 70526340)

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
sanjaiganesh: So this should always be Reject(new Error("error text"));

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
arjun-msft: I would love to do something like this but I think this should be the hub refactor of onboarding. That way when an access token is expired the hub knows how to authorize.  So what I'd suggest is the following: leave this the way it is, open an issue to track the right pattern to reuse code with hub and then refactor everything to follow the pattern. 

Thoughts?

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
arjun-msft: I'll update this but we need to have a better system of defining this.  init and disconnect don't have promise but the others do

org.OpenT2T.Sample.SuperPopular.Lamp/Vera Dimmable Light Bulb/js/thingTranslator.js
arjun-msft: What's the best way to validate 'this' problem?

jasongin: You can use [bind()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to explicitly set the value of `this` when calling a function. If you set it to the `global` object, that will mimic the current behavior of the AllJoyn-JXCore bridge. (Though that behavior should probably be considered a bug.)

jasongin:
👍

PR #33, opened 2016-07-18
Subscriptions

rlingineni:
I’m not exactly sure what to do when a change occurs, it currently just
prints to console if a change happens. 

Doorbell is very slow (maybe my internet connection, it’s slow with the Ring App too)

Checking in initial set of commits.

### DIFF
--- a/Helpers/OpenT2T-Wink-Helper/js/index.js
+++ b/Helpers/OpenT2T-Wink-Helper/js/index.js
@@ -227,7 +227,7 @@ module.exports =
         {
             try
             {    
-            //parse the JSON response and look for the apiField we want in the JSON body
+            //parse the JSON response and give back the pubnub details
              var results = JSON.parse(body.toString());
             deferred.resolve(results.data.subscription);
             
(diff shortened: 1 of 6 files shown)


rlingineni:
Maybe there can just be a `subscribeForStateChange()` in the schemas so it's clear they can subscribe to something. They should implement their own custom callback.

jasongin:
In the updated translator interface design I'm working on, any property can have the ability to notify of state changes. In the current format, those can be mapped to signals.

So I wouldn't worry about this right now; we'll update translators later for the new design.

tjaffri:
I'll take a look and merge. Sorry about the delay.

tjaffri:
just wanted to explain the delay in merging this one... in the dev branch, you will notice significant changes to our directory structure etc... specifically see #53.

Once we are done with those, we will likely need to rebase this PR to the new dir structure and package format. So please hang tight till then, and we will figure out how to merge this one.

arjun-msft:
Closing this out as we are leveraging PubSubHubBub for notifications.

PR #37, opened 2016-08-05
WINK Thermostat Translator that adopts the new design

sanjaiganesh:
@jasongin @tjaffri @arjun-msft @jeanpa Please review.


.gitignore
```diff
@@ -6,3 +6,5 @@ node_modules
 *.DS_Store
 logs
 *.log
+deviceconfig.js
+*.vscode
```
jasongin: Should this be deviceconfig.json ? JSON files can still be loaded directly as node modules.

jasongin: Actually, maybe a better name would be testconfig.json ?

---
In reply to: [73765456](https://github.com/openT2T/translators/pull/37#discussion_r73765456) [](ancestors = 73765456)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://standards.freedesktop.org/dbus/introspect-1.0.dtd"[]>
+<node>
+  <interface name="org.OpenT2T.Sample.SuperPopular.Thermostat">
+    <description>OpenT2T SuperPopular Thermostat interface</description>
+
+    <property name="PowerStatus" type="s" access="readwrite">
+      <description>Get/Set power status. (On, Off)</description>
+    </property>
+
```
jasongin: Should it be boolean?

Also, is this really separate from the Mode property? While I don't have experience with Wink, other smart thermostat APIs have included "Off" as one of the available modes.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "author": "OpenT2T",
+  "license": "MIT",
+  "scripts": {
+    "test": "ava"
+  },
+      "ava": {
+      "files": [
+        "tests/*.js",
+        "!**/deviceconfig.js"
+      ]
```
jasongin: This won't be needed if you give the file a .json extension.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+  "dependencies": {
+    "delay": "^1.3.1",
+    "opent2t": "*",
```
jasongin: This belongs in devDependencies, since it is only required by test code.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/ThermostatInterface.js
```diff
@@ -0,0 +1,6 @@
+// In a non-test interface module, this path would be: "opent2t/converters/AllJoynConverter"
+//var AllJoynConverter = require("../../../../build/lib/converters/AllJoynConverter");
+var AllJoynConverter = require("openT2T/converters/AllJoynConverter");
```
jasongin: Delete these lines to avoid confusion.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "opent2t-onboarding-winkhub": "*",
+    "opent2t-translator-helper-wink": "1.0.0",
+    "q": "1.4.1"
+  },
+  "devDependencies": {
+    "ava": "^0.15.2"
+  },
+  "opent2t": {
+    "interfaces": {
+      "org.OpenT2T.Sample.SuperPopular.Thermostat": {
```
jasongin: >org.OpenT2T.Sample.SuperPopular.Thermostat [](start = 7, length = 42)

This must match the name of the .js file, because it is the name of the interface module. The name of the .xml file doesn't really matter, though I had expected it would also be the same.

So either rename the .js file to org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.js, or change this to "ThermostatInterface". Or change both to just "Thermostat". Personally I'd recommend the latter.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "description": "OpenT2T SuperPopular Thermostat interface"
+      }
+    },
+    "translators": {
+      "TranslatorA": {
```
jasongin: >TranslatorA [](start = 7, length = 11)

This must match the name of the translator .js file, because it is the name of the translator module. I'd suggest naming them "WinkThermostat" and "WinkThermostat.js".

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "description": "OpenT2T Wink Thermostat translator",
+        "interfaces": [
+          "./ThermostatInterface"
```
jasongin: >ThermostatInterface [](start = 13, length = 19)

This must match the interface module name on line 28, whatever you change it to.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+        throw new Error('Invalid argument: ' + argName + '. ' +
+            'Expected type: ' + expectedType + ', got: ' + (typeof arg));
+    }
+}
+
+/**
+ * This translator class implements the "InterfaceA" interface.
+ * Since the interface includes a signal (notifiable property), this class
+ * extends from the node's built-in event-emitter class.
+ */
```
jasongin: Update comments.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+class WinkThermostat extends EventEmitter {
```
jasongin: >extends EventEmitter [](start = 21, length = 20)

EventEmitter is only needed for translators that emit signals. This one doesn't.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+
+    constructor(deviceInfo) {
+        super(); // Construct EventEmitter base
+        this.initDevice(deviceInfo);
```
jasongin: Also delete this when removing the base class.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    initDevice(deviceProps) {
+        console.log('Initializing device.');
```
jasongin: I'd suggest combining this method with the constructor.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+        console.log('Set mode.');
+        return wh.sendDesiredStateCommand('mode',mode); 
+    }
+
+    getAvailableModes() {
+        console.log('Get supported modes. But Allowed value depends on mode.choices in capabilities.');
+        var deferred = q.defer();   // Take a deferral
+       deferred.resolve(["cool_only", "heat_only", "auto", "aux"]);
+       return deferred.promise; // return the promise
+    }
```
jasongin: There's no need to use a promise. Just return the array result directly, and the DeviceAccessor will detect that it is an immediate result instead of a promise. Then you can remove the dependency on Q.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+        });
+    }
+    
+    setPowerStatus(status) {
+        console.log('Set PowerStatus.');
+        validateArgumentType(status, 'Power status', 'string');
+
+        if(status.toUpperCase() === 'ON')
+            return wh.sendDesiredStateCommand('powered', true);
+        else
```
jasongin: Eventually I'd suggest we set up [jslint](https://github.com/reid/node-jslint) to run on all our translator code, to keep them clean and consistently formatted. One of the things it will enforce is braces around conditional blocks.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
```diff
@@ -0,0 +1,32 @@
...
+    <property name="CurrentTemperature" type="d" access="read">
+     <description>Get the current ambient temperature in Celsius</description>
+    </method>
```
jasongin: >method [](start = 6, length = 6)

Change to </property>, here and below.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://standards.freedesktop.org/dbus/introspect-1.0.dtd"[]>
+<node>
+  <interface name="org.OpenT2T.Sample.SuperPopular.Thermostat">
+    <description>OpenT2T SuperPopular Thermostat interface</description>
+
+    <property name="PowerStatus" type="s" access="readwrite">
```
jasongin: >PowerStatus [](start = 20, length = 11)

Do we want to capitalize property names? I was thinking they would be camelCase, following JavaScript naming guidelines.

Note even when a property name in the interface is lowercase, if you implement it with a getter or setter it can be capitalized in context: powerStatus => getPowerStatus.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,117 @@
...
+
+var q = require('q');
+var wh = require("opent2t-translator-helper-wink");
+
+// This code uses ES2015 syntax that requires at least Node.js v4.
+// For Node.js ES2015 support details, reference http://node.green/
+
+const EventEmitter = require("events");
+
+function validateArgumentType(arg, argName, expectedType) {
```
jasongin: >validateArgumentType [](start = 9, length = 20)

I'm wondering if we should put stuff like this into the opent2t core library package, or maybe a separate "translator utils" package.

jasongin:
:clock1:

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
jasongin: Or we could declare that property accessors (and other methods) must *always* return promises. That would make it slightly easier for other JS code to consume translators more directly, at the expense of forcing translators to include silly code for returning completed promises like this.

---
In reply to: [73767459](https://github.com/openT2T/translators/pull/37#discussion_r73767459) [](ancestors = 73767459)

jasongin: (I think Q actually provides a more compact way to construct a completed promise, but I'd have to look it up.)

---
In reply to: [73770755](https://github.com/openT2T/translators/pull/37#discussion_r73770755) [](ancestors = 73770755,73767459)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
sanjaiganesh: I guess we should be following AllJoyn notation in the XML. corrrect ?

[https://wiki.allseenalliance.org/irb/interface_design_guidelines_1.0](https://wiki.allseenalliance.org/irb/interface_design_guidelines_1.0)

Member Names
Member (property, signal and method) names must use UpperCamelCase notation without punctuation

---
In reply to: [73768119](https://github.com/openT2T/translators/pull/37#discussion_r73768119) [](ancestors = 73768119)

jasongin: Oh, you're right. In that case, we'll need some code in DeviceAccessor to un-capitalize the names when looking for corresponding JavaScript translator properties, when the translator uses properties instead of accessor methods.

---
In reply to: [73771185](https://github.com/openT2T/translators/pull/37#discussion_r73771185) [](ancestors = 73771185,73768119)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: Ok. I shall retain this in that case. As you said, all properties are consistent in returning a promise.

jasongin: Umm, what I meant was we could require *all property accessors on all translators* to always return promises, not just in this class. I'm not sure I like the idea of imposing that requirement.

---
In reply to: [73773100](https://github.com/openT2T/translators/pull/37#discussion_r73773100) [](ancestors = 73773100)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
sanjaiganesh: Current DeviceAccessor expects get/set for properties. correct. I presume it is non-issue for now.

---
In reply to: [73772265](https://github.com/openT2T/translators/pull/37#discussion_r73772265) [](ancestors = 73772265,73771185,73768119)

.gitignore
sanjaiganesh: Cool. Fixed it.

---
In reply to: [73765465](https://github.com/openT2T/translators/pull/37#discussion_r73765465) [](ancestors = 73765465,73765456)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
sanjaiganesh: it is kind of wierd. Both are same. In case of power on, a default mode (heat-cool in case of Nest). Specific mode (heat/cool etc...) can be set separately.
Wink provides different property (powered) to turn on/off. But it ultimately sets hvac_mode. Nest has only hvac_mode property, which we are using in both setting mode and power on/off.

Regarding Boolean: I would make sense. I couldn't think of a proper property name. If it is 'Powered', methods will be getPowered and setPowered, which do not sound good.

---
In reply to: [73765703](https://github.com/openT2T/translators/pull/37#discussion_r73765703) [](ancestors = 73765703)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
sanjaiganesh: It is unclear whether it is file name or interface name. Will go for ThermostatInterface. Will retain the interface name to org.OpenT2T.... as that is AllJoyn standard.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
jasongin: DeviceAccessor currently supports both property accessors and actual JavaScript properties. Note that in JavaScript you can define code for property getters/setters, that are still invoked using property syntax, using [Object.defineProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty). And TypeScript has a nicer syntax for it, more like C# properties. Supporting that style of properties will make it easier for translator developers to use simple properties for things like device attributes and options.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
jasongin: It is really a "module name" for the interface. So it can be either a .js file name or the name of a directory that contains an index.js file for the module.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
jasongin: I'd call it "IsOn". But if it's totally redundant with the Mode property, and most thermostats don't have a notion of a separate power state, then I think we don't need it. Otherwise other kinds of thermostats that implement the interface would have trouble implementing it because they don't have a default mode.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: I have retained this, since Bridge calls into this routine, to init the device. It is unclear to me how the Bridge can consume this constructor pattern.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
sanjaiganesh: Removed PowerStatus properties

---
In reply to: [73777561](https://github.com/openT2T/translators/pull/37#discussion_r73777561) [](ancestors = 73777561)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: did setup on this.

---
In reply to: [73767716](https://github.com/openT2T/translators/pull/37#discussion_r73767716) [](ancestors = 73767716)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: yea. true.. we may have more such helper routines.. would defer that for now. Created an issue [https://github.com/openT2T/translators/issues/39](https://github.com/openT2T/translators/issues/39)


---
In reply to: [73768373](https://github.com/openT2T/translators/pull/37#discussion_r73768373) [](ancestors = 73768373)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: Created an issue to track for the repo [https://github.com/openT2T/translators/issues/38](https://github.com/openT2T/translators/issues/38)


---
In reply to: [73779631](https://github.com/openT2T/translators/pull/37#discussion_r73779631) [](ancestors = 73779631,73767716)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: @jasongin I would like to know your comment on this :  I have retained initDevice(), since Bridge calls into this routine, to init the device. It is unclear to me how the Bridge can consume this constructor pattern. 

---
In reply to: [73779241](https://github.com/openT2T/translators/pull/37#discussion_r73779241) [](ancestors = 73779241)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
sanjaiganesh: Changed to properties, instead of get/set accessor functions.

---
In reply to: [73776239](https://github.com/openT2T/translators/pull/37#discussion_r73776239) [](ancestors = 73776239)

sanjaiganesh: I had incorrectly assumed we always have to define get/set. 

---
In reply to: [73779790](https://github.com/openT2T/translators/pull/37#discussion_r73779790) [](ancestors = 73779790,73776239)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
jasongin: There is no point in trying to retain compatibility here with the existing bridge code. It already has no chance of working because of how the module exports are done differently: the class is exported rather than an object instance.

The bridge will soon (in a few weeks I guess) consume the OpenT2T client. Through the DeviceAccessor APIs projected through the client, it will have no trouble calling the constructor.

jasongin: There is no point in trying to retain compatibility here with the existing bridge code. It already has no chance of working because of how the module exports are done differently: the class is exported rather than an object instance.

The bridge will soon (in a few weeks I guess) consume the OpenT2T client. Through the DeviceAccessor APIs projected through the client, it will have no trouble calling the constructor.

---
In reply to: [73779700](https://github.com/openT2T/translators/pull/37#discussion_r73779700) [](ancestors = 73779700,73779241)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/org.OpenT2T.Sample.SuperPopular.Thermostat-2.0.xml
jasongin: Oh, nice, I didn't realize ES2015 supported that property syntax, I thought it was just a TypeScript thing.

However, I think we should allow developers to follow both JavaScript and AllJoyn naming guidelines. JavaScript property names should be camelCase. The DeviceAccessor can do the case conversion, as it already does in other situations.

---
In reply to: [73775104](https://github.com/openT2T/translators/pull/37#discussion_r73775104) [](ancestors = 73775104,73772265,73771185,73768119)

jasongin: Opened [https://github.com/openT2T/opent2t/issues/12](https://github.com/openT2T/opent2t/issues/12) to do the case conversion.

---
In reply to: [73780396](https://github.com/openT2T/translators/pull/37#discussion_r73780396) [](ancestors = 73780396,73775104,73772265,73771185,73768119)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
```diff
@@ -0,0 +1,88 @@
...
+    get HeatingSetpoint() {
+        console.log('Get heating point.');
+        return wh.getValueOfDesiredState('min_set_point');
+    }
+
+    set HeatingSetpoint(targetValue) {
+         console.log("Set Heating Setpoint");
+        return wh.sendDesiredStateCommand('min_set_point',targetValue);
+    }
+
```
jasongin: Unfortunately, this property declaration style is a problem when it comes to *async* setters, because set accessor methods can't a value (or if they do, the caller doesn't get it). But to be properly async, a set method like this needs to return a Promise that the caller can await. While I like the idea of using this syntax for properties, I think *async* properties are going to have to be normal getter and setter methods.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
jasongin: Agreed, I didn't intend for this to be fixed now.

---
In reply to: [73779662](https://github.com/openT2T/translators/pull/37#discussion_r73779662) [](ancestors = 73779662,73768373)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thermostat.js
```diff
@@ -0,0 +1,4 @@
+var AllJoynConverter = require("openT2T/converters/AllJoynConverter");
+
+module.exports = AllJoynConverter.readDeviceInterfacesFromFile(
+    require("path").join(__dirname, "thermostat.xml"))[0];
```
jasongin: >"thermostat.xml" [](start = 36, length = 16)

I think some people might prefer the name of the schema .xml to match the org.OpenT2T... name.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
jasongin: I guess this is OK for now, but it goes against the design principles that I had in mind for how we use modules. And having all translators named "thingTranslator.js" won't work when we transition to the new directory structure which can have multiple translators in a package.

---
In reply to: [73766739](https://github.com/openT2T/translators/pull/37#discussion_r73766739) [](ancestors = 73766739)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
jasongin: Somehow the wrong line got highlighted by my comment. The **delay** package belongs in devDependencies. The **opent2t** package belongs in depenencies, because it can be used at runtime if the interface is loaded for reflection.

---
In reply to: [73765884](https://github.com/openT2T/translators/pull/37#discussion_r73765884) [](ancestors = 73765884)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: Sure. All these wink helper routines (sendDesiredStateCommand)  return a promise. So should this be fine ?

---
In reply to: [73780491](https://github.com/openT2T/translators/pull/37#discussion_r73780491) [](ancestors = 73780491)

jasongin: Sorry if I wasn't clear: These need to be changed back to normal getter/setter methods.

There is an important semantic difference between these two declarations:
```
setHeatingSetpoint(targetValue) { ... }
set HeaintSetpoint(targetValue) { ... }
```
The first can return a value (a Promise in this case). For the second, the caller has no way to access the return value, because they invoke it like this:
```
translator.HeatingSetPoint = value;
```
Because of that, callers wouldn't be able to properly wait for async set operations.

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
sanjaiganesh: Sure. But even OpenT2T module is used only in testing,. 

---
In reply to: [73780663](https://github.com/openT2T/translators/pull/37#discussion_r73780663) [](ancestors = 73780663,73765884)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
sanjaiganesh: Sure. Will leave it like this for now. As I shall change this after new design if finalized. Jean and team has feedback and thoughts on the new design. There would be more changes.

---
In reply to: [73780613](https://github.com/openT2T/translators/pull/37#discussion_r73780613) [](ancestors = 73780613,73766739)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: Sure. Got it. Combined InitDevice with the constructor.

---
In reply to: [73780309](https://github.com/openT2T/translators/pull/37#discussion_r73780309) [](ancestors = 73780309,73779700,73779241)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/thingTranslator.js
sanjaiganesh: Ah .. sure. I could have thought through.

---
In reply to: [73826810](https://github.com/openT2T/translators/pull/37#discussion_r73826810) [](ancestors = 73826810)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
sanjaiganesh: sure. Makes sense.

---
In reply to: [73929203](https://github.com/openT2T/translators/pull/37#discussion_r73929203) [](ancestors = 73929203,73926071,73780663,73765884)

org.OpenT2T.Sample.SuperPopular.Thermostat/Wink Thermostat/v2/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "opent2t-translator-helper-wink": "1.0.0"
+  },
+  "devDependencies": {
+    "delay": "^1.3.1",
+    "jshint": "^2.9.2",
+    "ava": "^0.15.2"
+  },
+  "opent2t": {
+    "interfaces": {
+      "org.OpenT2T.Sample.SuperPopular.Thermostat": {
```
jasongin: >org.OpenT2T.Sample.SuperPopular.Thermostat [](start = 7, length = 42)

This is OK, but I was only suggesting to change the .xml file name, not necessarily the .js file name. The interface JavaScript file name (module name) can be simply "thermostat.js". The module name only needs to be unique within the package, not globally unique like the interface name.

sanjaiganesh: Sure. I had changed this, since Jean had similar feedback. I shall change it back to thermostat.js and hope to meet tomorrow to flush out final design.

---
In reply to: [73946648](https://github.com/openT2T/translators/pull/37#discussion_r73946648) [](ancestors = 73946648)

Issue #46, opened 2016-08-13
Add basic OCF schemas


tjaffri:
For #44 and #45 we need at least BinarySwitch and Dimming schemas from OCF. We should consider adding a few more as well, since they will likely be needed soon by translators we write.

arjun-msft:
This has been done.

Issue #44, opened 2016-08-13
Create org.OpenT2T.Sample.SuperPopular.Thermostat schema in OCF format


tjaffri:
Add json schema and RAML for org.OpenT2T.Sample.SuperPopular.Thermostat, using the OCF format.

See https://www.iotivity.org/

Issue #43, opened 2016-08-13
Update the directory structure of translators repo to use reverse URI based names for translators


tjaffri:
We currently have folders with the name of the hardware corresponding to each translator (without namespacing) under each schema folder.

 E.g. [here](https://github.com/openT2T/translators/tree/master/org.OpenT2T.Sample.SuperPopular.Lamp) we have folders like “Philips Hue Light Bulb” and “Osram Lightify Light Bulb”. These names are not namespaced, unlike schema names which are namespaced.

We need to update all the translator names to reverse URI based names, just like the schemas. The name of the hardware and the npm package name should be included in the package.json as description and name, respectively.

This will be a pretty disruptive change for anyone who depends on translators and the current structure of the repo, so we will need to make sure all dependencies are updated at the same time.

tjaffri:
We need some further changes here, based on discussion with @jasongin and @sanjaiganesh 

1. Move all reverse-uri naming schemes to lowercase (schema and translator folder names)
2. Tweak the names a little to use the "org.opent2t.sample.thermostat.superpopular" convention. This way they are still readable on GitHub but also following a naming hierarchy.

I will send another PR.

Thanks!

Issue #24, opened 2016-06-27
Add a signal for when the Doorbell button is pressed


tjaffri:
PR #23 added a translator and schema for a doorbell (with a concrete Wink doorbell implementation)

However, it is currently polling for whether the doorbell is pressed. This needs to be a signal that gets triggered when the doorbell is pressed.

tjaffri:
We can use this:

http://docs.wink.apiary.io/#reference/subscriptions

PR #48, opened 2016-08-14
Large dir structure changes, resolving #24 and #43

tjaffri:
@jeanpa @jasongin this is the first wave of repo dir structure changes we discussed. Would like feedback from everyone as well... @arjun-msft please check it out especially.

I will add on to this PR with the fixes for #44 and #46 but wanted to get some eyeballs on the changes as I go.

It is hard to view all these changes as a diff. All I can say is the changes are only moving things around... nothing functional changed. I tried to check to make sure I didn't miss anything but can fix in subsequent PRs if anything else needs to be updated.

Since this is a pretty large disruptive change I'd like to get this merged ASAP so we can review other changes to follo more easily and also start on fixing everything else that will break with these fixes (e.g. anything that depends on the repo structure, e.g. some of the sample apps)


jasongin:
I agree with Sanjai's comment (in email) that the reverse-dns names should be all lowercase.

jasongin:
What is the reason for switching the order of the schema name components, from "org.opent2t.superpopular.thermostat" to "org.opent2t.thermostat.superpopular.sample"?

Also I don't understand why we need both "superpopular" and "sample" in the names. So I would suggest names like "org.opent2t.sample.thermostat".

jasongin:
:clock1:

tjaffri:
@jasongin and @sanjaiganesh I discussed with @jeanpa about both of these, and they are #ByDesign. Specifically:

1. Switching the name of the schema name components from "org.opent2t.superpopular.thermostat" to "org.opent2t.thermostat.superpopular.sample" is for readability on GitHub... so we can actually read what translator is what when you go to the repo home on GitHub.

2. The capitalization is also something we have done since the beginning (see original schema names). We can certainly reconsider going forward, but for now not changing that pre-exisiting convention.

tjaffri:
Also, @jeanpa gave a verbal :shipit: so merging on his behalf

Issue #45, opened 2016-08-13
Modify the Wink Thermostat Translator to be aligned with OCF format


tjaffri:
#44 is creating the thermostat schema in OCF format.

In parallel, we need to update one translator (let's pick the Wink Thermostat Translator for ease) to implement that schema. This includes:

## Updating the manifest file
The `manifest.xml` needs to reference the OCF schema(s), where it currently only references a single AllJoyn schema.

For example:

```xml
<schemas>
      <schema id="org.opent2t.sample.thermostat.superpopular" main="true" type="OCF" />
      <schema id="oic.r.temperature" type="OCF" />
      <schema id="oic.core" type="OCF" />
      <schema id="oic.baseResource" type="OCF" />
  </schemas>
```

Note that we are OK (for now) duplicating the fact the dimming and binary will likely be $ref inside org.OpenT2T.Sample.SuperPopular.Lamp but will also be duplicated under schemas. We can discuss this further at a later date.

## Updating the methods in the translator itself
We need to update the translator methods to correspond to the RAML (get/put pattern) in OCF.

tjaffri:
@jasongin and @sanjaiganesh I was doing some other dir structure renaming work with @arjun-msft and @jeanpa and we STUBBED OUT some of this stuff. I haven't made all the changes like updating the test automation and the actual deviceTranslator implementation but let's huddle to see if you can reuse some of my stub.

jasongin:
We need to clarify the usage of `<schema>` elements in the manifest.
* The first element in the example above does not include a filename extension -- does it refer to the .raml or .json file?
* A .raml file may reference more than one top-level .json schema, either because it defines multiple endpoints or because different verbs for the same endpoint use different schemas. (Though it appears neither of those are common practice for OCF, I wouldn't want the design to rely on that assumption.) Should each of them be listed separately in the manifest?
* What about non-top-level .json schema references, should they be listed also?

In my opinion, the manifest only needs to reference the .raml file. I don't see any reason to duplicate all of the direct and indirect references to .json schema files (a flattened schema dependency tree) in the manifest.

tjaffri:
Here are some notes on the `<schema>` elements in the manifest:

1. Really all we need in the manifest is a way to locate the .js file in the manifest directory, specifically the name of the main schema. So in the given example all we need is `org.opent2t.sample.thermostat.superpopular`

[...]

Issue #51, opened 2016-08-16
Create the script to package a translator into an npm package


tjaffri:
@jasongin you said you are already working on this, so tracking with an issue.

tjaffri:
We should also think about how this impacts the files: attribute in the package.json. Perhaps auto-generate it based on which files are included in the package?

Issue #36, opened 2016-08-02
Update all translators to new design


tjaffri:
Update translators for breaking changes in the new design, including:

  - Node module exports
  - Translator metadata in package.json
  - Schema composability, and directory structure changes that result from it
  - Properties and signals
  - Hub devices

tjaffri:
We should spin up a separate branch to do these changes, and coordinate across repos.

tjaffri:
This wiki contains more information about the new design: https://github.com/openT2T/opent2t/wiki/Modules-and-Packages

sanjaiganesh:
@tjaffri @arjun-msft Isn't this is something to be distributed across all of us, once Wink Thermostat with new design PR is merged. ?

arjun-msft:
This issue is now out of date.

PR #53, opened 2016-08-17
Large set of changes (see details)

### LINKED ISSUE #43: Update the directory structure of translators repo to use reverse URI based names for translators

### LINKED ISSUE #44: Create org.OpenT2T.Sample.SuperPopular.Thermostat schema in OCF format

### LINKED ISSUE #46: Add basic OCF schemas

### LINKED ISSUE #50: Node package naming conventions
Every translator has a node package name in its package.json, e.g. 

```json
{
    "name": "opent2t-translator-wink-light-bulb",
    "version": "1.0.0",
    "description": "Wink Light Bulb",
    "homepage": "http://opentranslatorstothings.org",
    "dependencies": {},
    "devDependencies": {
        "async": "~1.5.2",
        "optimist": "0.6.1",
        "opent2t-onboarding-winkhub": "*"
    },
    "author": "",
    "license": "MIT"
}
```

We need to come up with a convention for these names, somehow aligning with the reverse-URI naming convention of the translator folders... This needs to get done before we publish these to NPM, but since it is not a directory structure change we can separate this from #43

### LINKED ISSUE #52: Wink thermostat translator fails with a redirect to the Wink homepage
Running the test automation recently introduced [here]() the Wink translator is failing, and redirects to the wink home page.

We need to figure out why. Perhaps we need to move to the v2 API sooner than we thought?

tjaffri:
There are a few large fixes in here. See below:

1. Resolves #50 - Node package naming conventions. All the package.json files are now in this format.
2. Resolves #44 - Thermostat schema in OCF format after many discussions across @jeanpa, @arjun-msft and @jasongin. Thanks everyone for your help!
3. Resolves #46 - Basic OCF schemas. These are checked in under the oic.r.* namespaces, and are referenced from the Thermostat schema.
4. Partial fixes for #45 - Thermostat translator in OCF format. The Get and Post are currently stubbed out, but I need to fix #52 before I implement these. Wanted to get this PR merged first.
5. Resolves #43 - reverse URI based names for translators. Thanks everyone for the feedback in that regard.

Sorry this is such a large merge... but needed multiple rounds of feedback of the holistic design. Will try to make future PRs atomic and easier to review one at a time.


org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular.xml
```diff
@@ -0,0 +1,24 @@
...
+    </property>
+
+    <property name="targetTemperatureHigh" type="d" access="readwrite">
+      <description>Maximum target temperature (in Celsius).</description>
+    </property>
+
+    <property name="targetTemperatureLow" type="d" access="readwrite">
+      <description>Minimum target temperature (in Celsius).</description>
+    </property>
+
```
sanjaiganesh: change units to fahrenheit

sanjaiganesh:
>         "optimist": "0.6.1",

Are we not adding 'files' field ? Is it not required ?

---
Refers to: org.opent2t.sample.windowshade.superpopular/org.opent2t.test.windowshade/js/package.json:10 in 0f69203. [](commit_id = 0f692033669da96e00df195201b5162070128e1d, deletion_comment = False)

sanjaiganesh:
:clock1:

tjaffri:
@sanjaiganesh per the discussion in https://github.com/openT2T/translators/issues/51 we should put that in the build script rather than include files inside the package.json as a fixed entry. Please file a separate tracking issue if you are concerned.

org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular.xml
tjaffri: Actually, I stuck with Celsius on purpose.... that's the SI unit. May be a little un-american of me, perhaps :)

For what it's worth, the Wink API (for which we are writing the first translator) also uses Celsius units: http://docs.winkapiv2.apiary.io/#reference/device/thermostats

#ByDesign

jasongin: I agree with using Celsius units here, as long as the type is decimal not integer so there is no loss of precision when converting on either side.

Note the OCF thermostat schema that we've been discussing actually specifies the units in the property JSON.

(I am an American, and I think non-metric units are silly and old-fashioned.)

jasongin:
>         "optimist": "0.6.1",

Any use of "files" or "main" in the package.json here is going to be wrong because of how the package.json file is in the wrong directory. Both properties will be automatically generated/overwritten by the build script anyway.

---
In reply to: [240483336](https://github.com/openT2T/translators/pull/53#issuecomment-240483336) [](ancestors = 240483336)

---
Refers to: org.OpenT2T.WindowShade.SuperPopular.Sample/org.OpenT2T.Test.WindowShade/js/package.json:10 in 0f69203. [](commit_id = 0f692033669da96e00df195201b5162070128e1d, deletion_comment = False)

org.opent2t.sample.thermostat.superpopular/com.nest.thermostat/js/manifest.xml
```diff
@@ -3,9 +3,9 @@
   <!-- associated voice handler -->
   <voice id="" />
   <!-- associated schema -->
-  <schema id="org.OpenT2T.Thermostat.SuperPopular.Sample" />
+  <schema id="org.opent2t.sample.thermostat.superpopular" />
   <!-- associated onboarding module -->
```
jasongin: Are we intending to support both a single <schema> element at the top level as well as one or more <schema> elements under a <schemas> element? The parsing code in my lib PR currently only supports the latter, but I can change it.

jasongin:
> 

This class follows the AllJoyn schema, which is not in sync with the RAML schema of the same name. Since the schema .js file references the RAML version, this translator is not consistent with that.

---
Refers to: org.opent2t.sample.thermostat.superpopular/com.nest.thermostat/js/thingTranslator.js:91 in 0f69203. [](commit_id = 0f692033669da96e00df195201b5162070128e1d, deletion_comment = False)

org.opent2t.sample.thermostat.superpopular/com.wink.thermostat/js/tests/test.js
```diff
@@ -0,0 +1,65 @@
+/*jshint esversion: 6 */
+var opent2t = require('opent2t');
+var ava = require('ava');
+var config = require('./testconfig');
+var delay = require('delay')
+
+// This boilerplate code is copied from transpiled typescript AVA test code, which enables usage of 'yield' keyword. 
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
jasongin: I'm concerned that this, and the use of __awaiter and yield below, makes the test code look far more complicated than it needs to be. We want other developers to be able to copy this test code to use for writing tests for translators they develop. If we're not using TypeScript, then we should just write the test code in normal ES6 syntax with .then() callbacks instead of awaits.

oic.baseResource/oic.baseResource.json
```diff
@@ -0,0 +1,61 @@
+{
+    "id": "http://openinterconnect.org/iotdatamodels/schemas/oic.baseResource.json#",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "Copyright (c) 2016 Open Connectivity Foundation, Inc. All rights reserved.",
+    "title": "Base Resource",
+    "definitions": {
+        "oic.r.baseresource": {
```
jasongin: >oic.r.baseresource [](start = 9, length = 18)

There is actually a "baseResource" RAML at oneiota.org, should that be included here?

(I didn't find a "core" RAML.)

org.opent2t.sample.thermostat.superpopular/com.wink.thermostat/js/tests/test.js
sanjaiganesh: This is .js file. I felt this was looking clean and we really want us to use a test runner like ava, for all its benefits. when we setup lab, we know how many total test cases and how many passed etc.. May be we can simplify this by removing this boilerplate code and fixing it.

---
In reply to: [75198452](https://github.com/openT2T/translators/pull/53#discussion_r75198452) [](ancestors = 75198452)

org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular.xml
sanjaiganesh: so we should fix the schemas  to "units": "C" ?

---
In reply to: [75194786](https://github.com/openT2T/translators/pull/53#discussion_r75194786) [](ancestors = 75194786)

org.opent2t.sample.thermostat.superpopular/com.wink.thermostat/js/tests/test.js
jasongin: We can have all those benefits while using normal .then() callbacks instead of this __awaiter which is messy and unfamiliar to people. AVA doesn't require TypeScript or await.

---
In reply to: [75208579](https://github.com/openT2T/translators/pull/53#discussion_r75208579) [](ancestors = 75208579,75198452)

org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular.xml
jasongin: There's not much point in making changes to the AllJoyn schemas now, until we decide whether and how we are going to align the AllJoyn schemas with OCF. Maybe we will even auto-convert them from OCF.

---
In reply to: [75208998](https://github.com/openT2T/translators/pull/53#discussion_r75208998) [](ancestors = 75208998,75194786)

tjaffri: @sanjaiganesh I think this is #ByDesign. If you disagree, please file a bug.

org.opent2t.sample.thermostat.superpopular/com.nest.thermostat/js/manifest.xml
tjaffri: This nest thermostat translator has not been updated as part of this PR (none of the other translators have been updated either). So let's ignore whatever that says (other than renaming/casing changes of course).

The Wink Thermostat is under review here. Issue #36 tracks updating all the other translators (Once we agree on this one I can do a series of PRs modifying all the other translators one by one following the same pattern)

oic.baseResource/oic.baseResource.json
tjaffri: Thanks! I'll add the RAML in a subsequent PR. Added this issue to track adding that later: https://github.com/openT2T/translators/issues/54 so for now this is #WontFix in this PR (hope I got that terminology right?)

org.opent2t.sample.thermostat.superpopular/com.wink.thermostat/js/tests/test.js
tjaffri: This is a great discussion, but I'm hoping we don't need to block this PR for this. @sanjaiganesh I added issue #55 to track this... perhaps you could take a stab at it? For this PR I think this is #WontFix and we can clean it up later (very soon hopefully, perhaps tomorrow/friday).

tjaffri:
@jasongin as long as you see my note about the nest thermostat translator being out of scope for this PR, I think I have addressed all feedback so far.

Do you guys have more feedback?

jasongin:
:shipit:

sanjaiganesh:
:shipit:

PR #63, opened 2016-08-22
Test fix for the Wink thermostat

### LINKED ISSUE #62: Unable to change Wink device properties after new schema changes
After integrating with the new schema changes I can get the Wink device properties from the translator with getThermostatResURI but i'm unable to change the device properties using postThermostatResURI . The post succeeds but the device does not update.

tjaffri:
Sets the mode and then queries the corresponding setpoint.

This is related to #62 but should only be a reference. To fix #62 we need to make the corresponding changes in the translator, schema, unit tests and any callers as discussed in that issue...


tjaffri:
test-only fix, so merging.

PR #65, opened 2016-08-23
Fixing bug for setting min/max set point

arjun-msft:
Fix for issue https://github.com/openT2T/translators/issues/62


tjaffri:
:shipit:

Issue #61, opened 2016-08-22
Complete the Thermostat schema, and Wink Thermostat Translator


tjaffri:
#44 and #45 scaffolded out the initial pattern for Thermostat schema, with a basic implementation for the Wink Thermostat Translator (in the dev branch, at the time of writing).

We now need to complete the implementation. Specifically:

1. Add more schema properties, specifically things like thermostat modes etc.
2. Implement the new schema methods in the translator.

For both above, we can start with the OCF style schemas (RAML and JSON) with corresponding OCF style translator implementation (get and post methods). However, to completely close this issue we need to (later) implement the AllJoyn style schema as well.

Finally, we should include automated tests as well. See the tests folder inside the existing Wink Thermostat translator for some examples. We should extend for new schema properties.

> **Note:** All wink translators depend deeply on opent2t-translator-helper-wink, which is checked into the Helpers folder in this repo. We will move this code elsewhere soon, but for now you will need to use this helper (and add to it / fix it if needed) to resolve this issue.

PR #69, opened 2016-08-23
Update thermostat schema and complete Wink thermostat translator

begillil:
The changes for #61 to complete the Wink thermostat translator and update the thermostat schema


tjaffri:
:shipit: I have some refactoring suggestions to put your helpful schema reader etc in a common place where other wink translators can use them... but I will file issues for those so we can move fast (and also because @arjun-msft is working on getting rid of the helper and putting common methods elsewhere so he can better advise).

PR #73, opened 2016-08-24
Adding eslint to nest translator

ChuckFerring:
Adding eslint to a single translator.  Once any feedback is addressed I'll update the other translators and the translator bootstrapping scripts as well.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,3 @@
+{
+    "extends": "opent2t"
+}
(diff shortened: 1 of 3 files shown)


tjaffri:
@ChuckFerring two suggestions:

1. Could you please update the README as well with the one or two lines of bash required to lint the translator? See the testing instructions for reference... I assume it is 'npm lint' but let's have that handy for future reference.
2. Maybe we should lint as part of the 'npm test' command? or perhaps have an 'npm build' command or similar that does both linting and testing? I don't know what the correct pattern/practice here so please advise.

arjun-msft:
:shipit:

Issue #40, opened 2016-08-10
Update Wink translators to v2 API


tjaffri:
http://docs.winkapiv2.apiary.io/#

tjaffri:
Looks like we need a special client ID and client secret to get the v2 API. Need to contact Wink to get it.. till then we are technically still on the v1 API.

PR #75, opened 2016-08-24
Initial checkin of hub schema and wink hub translator.

arjun-msft:
This is part 1 of the complete change, this is the hub schema and the wink hub translator.  Part 2 will involve removing the wink helper and updating the wink translators to use the hub instead of the helper.

### DIFF
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
@@ -1,23 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest>
-  <!-- associated voice handler -->
-  <voice id="org.opent2t.sample.lightcontrol.cortana.vcd.superpopular" />
-  <!-- associated schemas -->
   <schemas>
-      <schema id="org.opent2t.sample.thermostat.superpopular" main="true" type="OCF" />
-      <schema id="oic.r.temperature" type="OCF" />
+      <schema id="org.opent2t.sample.hub.superpopular" main="true" type="OCF" />
       <schema id="oic.core" type="OCF" />
       <schema id="oic.baseResource" type="OCF" />
   </schemas>
   <!-- associated onboarding module -->
   <onboarding id="org.opent2t.onboarding.winkhub">
     <onboardflow name="getUserInput">
-      <flow type=”input”>
+      <flow type="input">
         <arg name="username" />
         <description language="en">Type in your Wink username </description>
         <description language="fr">Entrez for nom pour Wink </description>
       </flow>
-      <flow type=”password”>
+      <flow type="password">
         <arg name="password"  />
         <description language="en">Type in your Wink password </description>
         <description language="fr">Entrez for password pour Wink </description>
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest>
+  <!-- associated voice handler -->
+  <voice id="org.opent2t.sample.lightcontrol.cortana.vcd.superpopular" />
+  <!-- associated schemas -->
+  <schemas>
+      <schema id="org.opent2t.sample.thermostat.superpopular" main="true" type="OCF" />
+      <schema id="oic.r.temperature" type="OCF" />
+      <schema id="oic.core" type="OCF" />
+      <schema id="oic.baseResource" type="OCF" />
+  </schemas>
+  <!-- associated onboarding module -->
+  <onboarding id="org.opent2t.onboarding.winkhub">
+    <onboardflow name="getUserInput">
+      <flow type=”input”>
+        <arg name="username" />
+        <description language="en">Type in your Wink username </description>
+        <description language="fr">Entrez for nom pour Wink </description>
+      </flow>
+      <flow type=”password”>
+        <arg name="password"  />
+        <description language="en">Type in your Wink password </description>
+        <description language="fr">Entrez for password pour Wink </description>
+      </flow>
+    </method>
+
+    <onboardflow name="getDeveloperInput">
+      <flow>
+        <arg name="client_id" />
+        <description language="en"> Ask for Client ID </description>
+        <description language="fr"> Demandez Client ID </description>
+      </flow>
+      <flow>
+        <arg name="client_secret" />
+        <description language="en"> Ask for Client Secret </description>
+        <description language="fr"> Demandez Client Secret </description>
+      </flow>
+    </onboardflow>
+
+  </onboarding>
+</manifest>
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "opent2t-wink-hub",
+  "version": "1.0.0",
+  "description": "Hub schema and helper modules for Open Translators to Things",
+  "author": "OpenT2T",
+  "license": "MIT",
+  "main": "thingTranslator.js",
+  "files": [
+    "thingTranslator.js",
+    "package.json"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "githttps://github.com/openT2T/translators.git"
+  },
+  "bugs": {
+    "url": "https://github.com/openT2T/translators/issues"
+  },
+  "homepage": "https://github.com/openT2T/translators#readme",
+  "dependencies": {
+    "request": "^2.74.0",
+    "request-promise": "^4.1.1"
+  },
+  "devDependencies": {
+    "opent2t": "*",
+    "ava": "^0.15.2"
+  }
+}
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
@@ -1,8 +1,7 @@
 {
-  "name": "opent2t-wink-hub",
+  "name": "opent2t-translator-com-wink-hub",
   "version": "1.0.0",
-  "description": "Hub schema and helper modules for Open Translators to Things",
-  "author": "OpenT2T",
+  "description": "Wink Hub",
   "license": "MIT",
   "main": "thingTranslator.js",
   "files": [
@@ -19,10 +18,10 @@
   "homepage": "https://github.com/openT2T/translators#readme",
   "dependencies": {
     "request": "^2.74.0",
-    "request-promise": "^4.1.1"
+    "request-promise": "^4.1.1",
+    "opent2t": "^1.0.1"
   },
   "devDependencies": {
-    "opent2t": "*",
     "ava": "^0.15.2"
   }
 }
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
@@ -1,29 +1,37 @@
+// var schema = require("c:/githome/translators/org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular");
+// console.log(schema);
+
 var test = require('ava');
 var OpenT2T = require('opent2t').OpenT2T;
 var config = require('./testConfig');
 
-console.log(config);
-//console.log('Device Under Test -  Name: ' + config.Device.name + '  Props: ' + JSON.stringify(config.Device.props));
+console.log("Config:");
+console.log(JSON.stringify(config, null, 2));
 var translatorPath = require('path').join(__dirname, '..');
 
 ///
 /// Run a series of tests to validate the translator
 ///
 
-// Get AmbientTemperature
-test.serial('GetDevices', t => {
+// HubResURI
+test.serial('HubResURI', t => {
 
     return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config)
         .then(translator => {
             // TEST: translator is valid
             t.is(typeof translator, 'object') && t.truthy(translator);
-            return OpenT2T.getPropertyAsync(translator, 'org.opent2t.sample.hub.superpopular', 'devices')
-                .then((devices) => {
+            return OpenT2T.getPropertyAsync(translator, 'org.opent2t.sample.hub.superpopular', 'HubResURI')
+                .then((hub) => {
 
-                    // TEST: some ambient temperature was returned
-                    console.dir(devices);
-                    t.truthy(devices);
-                    t.true(devices.length > 0);
+                    console.log("Hub:");
+                    console.log(JSON.stringify(hub, null, 2));
+
+                    // TEST: something was returned
+                    t.truthy(hub);
+
+                    // TEST: hub has devices
+                    t.truthy(hub.devices);
+                    t.true(hub.devices.length > 0);
                 });
         });
 });
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
@@ -1,0 +1,29 @@
+var test = require('ava');
+var OpenT2T = require('opent2t').OpenT2T;
+var config = require('./testConfig');
+
+console.log(config);
+//console.log('Device Under Test -  Name: ' + config.Device.name + '  Props: ' + JSON.stringify(config.Device.props));
+var translatorPath = require('path').join(__dirname, '..');
+
+///
+/// Run a series of tests to validate the translator
+///
+
+// Get AmbientTemperature
+test.serial('GetDevices', t => {
+
+    return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config)
+        .then(translator => {
+            // TEST: translator is valid
+            t.is(typeof translator, 'object') && t.truthy(translator);
+            return OpenT2T.getPropertyAsync(translator, 'org.opent2t.sample.hub.superpopular', 'devices')
+                .then((devices) => {
+
+                    // TEST: some ambient temperature was returned
+                    console.dir(devices);
+                    t.truthy(devices);
+                    t.true(devices.length > 0);
+                });
+        });
+});
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
@@ -23,9 +23,11 @@ class Translator {
     /**
      * Get the list of devices discovered through the hub.
      */
-    getDevicesAsync() {
-        return this._makeRequest(this._devicesPath, 'GET').then((data) => {
-            var devices = data.data;
+    getHubResURI() {
+        return this._makeRequest(this._devicesPath, 'GET').then((response) => {
+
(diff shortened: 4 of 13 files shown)


org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
```diff
@@ -0,0 +1,41 @@
...
+  </schemas>
+  <!-- associated onboarding module -->
+  <onboarding id="org.opent2t.onboarding.winkhub">
+    <onboardflow name="getUserInput">
+      <flow type=”input”>
+        <arg name="username" />
+        <description language="en">Type in your Wink username </description>
+        <description language="fr">Entrez for nom pour Wink </description>
+      </flow>
+      <flow type=”password”>
```
begillil: ><flow type=”input”> 
><flow type=”password”> 

The quotation marks for these look odd. Perhaps they're using the wrong quote symbol?

org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest>
+  <!-- associated voice handler -->
+  <voice id="org.opent2t.sample.lightcontrol.cortana.vcd.superpopular" />
+  <!-- associated schemas -->
+  <schemas>
+      <schema id="org.opent2t.sample.thermostat.superpopular" main="true" type="OCF" />
```
tjaffri: this should be org.opent2t.sample.hub.superpopular

org.opent2t.sample.hub.superpopular/com.wink.hub/js/manifest.xml
```diff
@@ -0,0 +1,41 @@
...
+      <schema id="oic.r.temperature" type="OCF" />
```
tjaffri: This shouldn't be here, since the hub schema should not reference temperature (only core and baseResource should be sufficient)

begillil:
So, device enumeration is done through a 'hub' device, which maps to enumerating all devices associated with a Wink account.

How do we handle a Wink account with multiple hubs attached to it?

org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "opent2t-wink-hub",
```
tjaffri: We are using a different pattern... see https://github.com/openT2T/translators/issues/50. I believe this should be opent2t-translator-com-wink-hub

We should probably doc this somewhere... I opened #50 to track.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.0.0",
+  "description": "Hub schema and helper modules for Open Translators to Things",
+  "author": "OpenT2T",
```
tjaffri: None of the other translators have this... I suggest deleting. Also, other translators have he "homepage" element. I suggest referencing https://github.com/openT2T/translators/blob/dev/org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/package.json and following as closely as possible.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "opent2t-wink-hub",
+  "version": "1.0.0",
+  "description": "Hub schema and helper modules for Open Translators to Things",
```
tjaffri: For consistency with other translators, I suggest this should just be "Wink Hub". We are using this field as the "friendly human readable name of this thing".

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.json
```diff
@@ -0,0 +1,24 @@
...
+  "definitions": {
+    "org.opent2t.sample.hub.superpopular": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "ReadOnly, Gets the name of the hub."
+        },
+        "devices": {
+          "description": "ReadOnly, Gets the current set of devices associated with the hub."
+        }
```
begillil: I believe these properties should have some sort of type.

Also, oic.core defines a friendly name field already: 'n', so I don't believe you need another one.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
```diff
@@ -0,0 +1,29 @@
+var test = require('ava');
+var OpenT2T = require('opent2t').OpenT2T;
+var config = require('./testConfig');
+
+console.log(config);
```
tjaffri: Do we need to JSON stringify? This just prints out as [object, object] for me... don't know why? Recommend you use the commented out line below or modify it ... might want to do JSON.stringify(foo, null, 2) to make it pretty per earlier suggestion by @begillil.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/tests/test.js
```diff
@@ -0,0 +1,29 @@
...
+//console.log('Device Under Test -  Name: ' + config.Device.name + '  Props: ' + JSON.stringify(config.Device.props));
+var translatorPath = require('path').join(__dirname, '..');
+
+///
+/// Run a series of tests to validate the translator
+///
+
+// Get AmbientTemperature
```
tjaffri: Fix... should be Get Devices

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+        this._baseUrl = "https://api.wink.com";
+        this._devicesPath = '/users/me/wink_devices';
+
+        this._name = "Wink Hub"; // TODO: Can be pulled from OpenT2T global constants. This information is not available, at least, on wink hub.
+    }
+
+    /**
+     * Get the list of devices discovered through the hub.
+     */
+    getDevicesAsync() {
```
tjaffri: Are you sure this Async prefix is supported? I just do get[PropertyName] so in this case I would have done getDevices().  See here for [example](https://github.com/openT2T/translators/blob/dev/org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/thingTranslator.js#L122). Maybe this is a bug in my translators? @jasongin what's the correct pattern here?

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.raml
```diff
@@ -0,0 +1,36 @@
...
+schemas:
+  - Hub: !include org.opent2t.sample.hub.superpopular.json
+
+traits:
+  - interface:
+      queryParameters:
+        if:
+          enum: ["oic.if.a"]
+
+/hubResURI:
```
tjaffri: This should start with a capital letter for consistency (I think) with the OCF stuff. See [here](https://github.com/openT2T/translators/blob/dev/oic.r.mode/mode.raml) and [here](https://github.com/openT2T/translators/blob/dev/oic.r.switch.binary/binarySwitch.raml) as two examples.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+     * Get the name of the hub.
+     */
+    getName() {
+        return this._name;
+    }
+
+    /**
+     * Gets device details (all fields), response formatted per http://docs.winkapiv2.apiary.io/
+     */
+    getDeviceDetailsAsync(deviceType, deviceId) {
```
tjaffri: Where is this used?

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+                console.log("Error            : " + err.statusCode + " - " + err.response.statusMessage);
+                // todo auto refresh in specific cases, issue 74
+                throw err;
+            });
+    }
+
+    /**
+     * Wink has a semi-complicated way of getting the actual device id of a device.
+     */
+    _getDeviceId(winkDevice) {
```
tjaffri: Just as a general note (not a problem for this PR) this is specific to the v2 API. See issue #40 and after that we should not need to do this assuming the caller is on the v2 API.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+// This code uses ES2015 syntax that requires at least Node.js v4.
+// For Node.js ES2015 support details, reference http://node.green/
+
+"use strict";
+var request = require('request-promise');
+
+/**
+* This translator class implements the "Hub" interface.
+*/
+class Translator {
```
tjaffri: Looks like this translator is not implementing the OCF style getHubResURI and postHubResURI(value) methods? We should since that's a pattern (see the nest and wink thermostat translators for reference)

tjaffri:
🕐

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.xml
```diff
@@ -0,0 +1,16 @@
...
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://standards.freedesktop.org/dbus/introspect-1.0.dtd"[]>
+<node>
+  <interface name="org.opent2t.sample.hub.superpopular">
+    <description>OpenT2T SuperPopular Hub interface</description>
+
+    <property name="name" type="s" access="read">
+     <description>ReadOnly, Gets the name of the hub.</description>
+    </property>
+    
+    <property name="devices" type="a" access="read">
```
jasongin: >a [](start = 35, length = 1)

Just "a" by itself isn't a valid AllJoyn type code. It has to be an array of *something*.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "url": "https://github.com/openT2T/translators/issues"
+  },
+  "homepage": "https://github.com/openT2T/translators#readme",
+  "dependencies": {
+    "request": "^2.74.0",
+    "request-promise": "^4.1.1"
+  },
+  "devDependencies": {
+    "opent2t": "*",
+    "ava": "^0.15.2"
```
jasongin: This should be in dependencies, not devDependencies, because the schema has a runtime dependency on the schema reader in this package. Also specify a version, not just "*".

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
jasongin: Yes the Async suffix is supported (but optional) for property getters/setters and any methods. But as mentioned above, this should be refactored to OCF style methods.

---
In reply to: [76154581](https://github.com/openT2T/translators/pull/75#discussion_r76154581) [](ancestors = 76154581)

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+            });
+    }
+
+    /**
+     * Wink has a semi-complicated way of getting the actual device id of a device.
+     */
+    _getDeviceId(winkDevice) {
+        // get the device id
+        var deviceId = winkDevice.light_bulb_id ||
+            winkDevice.air_conditioner_id ||
```
jasongin: It looks like this list could easily be subject to change. Consider enumerating `Object.keys(winkDevice)` and looking for a property that has the "_id" suffix.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -0,0 +1,187 @@
...
+
+        // Make the async request
+        return _makeRequest(requestPath, 'PUT', putPayloadString);
+    }
+
+    /** 
+     * Given the hub specific device, returns the opent2t schema and translator
+    */
+    _getOpent2tInfo(winkDevice) {
+        if (winkDevice.thermostat_id != undefined) {
```
jasongin: >!= [](start = 37, length = 2)

The linter will catch this: use `!==` instead of `!=`. And yes, the difference is significant:

```
null != undefined   // false
null !== undefined  // true
```

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
begillil: It should be noted that there are caveats involved with that approach. I believe every Wink device has a hub_id, not just hubs, and there may be other cases where this doesn't work out.

Fortunately, the V2 APIs have an object_type field which makes this problem easier to solve when we transition fully.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
jasongin: Or maybe what you really wanted was to just check if the id is *truthy*:

```
if (winkDevice.thermostat_id) {
```

---
In reply to: [76161980](https://github.com/openT2T/translators/pull/75#discussion_r76161980) [](ancestors = 76161980)

arjun-msft:
This feedback is awesome, I'll work on fixing this and sending an updating PR.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
arjun-msft: This get/put are the helper methods which the wink devices will use.

---
In reply to: [76154976](https://github.com/openT2T/translators/pull/75#discussion_r76154976) [](ancestors = 76154976)

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
arjun-msft: Correct, this is just a v1 issue, will be resolved in v2.

---
In reply to: [76162061](https://github.com/openT2T/translators/pull/75#discussion_r76162061) [](ancestors = 76162061)

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.json
begillil: 'devices' still needs a type.

---
In reply to: [76154047](https://github.com/openT2T/translators/pull/75#discussion_r76154047) [](ancestors = 76154047)

arjun-msft: I'm not sure what to do here, should I create a custom type? would love your guidance/recommendation. the data that gets returned is: 
  "devices": [
    {
      "name": "Home Kitchen Thermostat (AB92)",
      "id": "152846",
      "openT2T": {
        "schema": "org.opent2t.sample.thermostat.superpopular",
        "translator": "com-wink-thermostat"
      }
    }

This is per the apiary doc. 

---
In reply to: [76334693](https://github.com/openT2T/translators/pull/75#discussion_r76334693) [](ancestors = 76334693,76154047)

arjun-msft:
We need to design that in still.  There is conversation around creating a higher level concept of provider/account which enumerates the entire tree of devices. 

We should include you in future discussions around this.

---
In reply to: [242231119](https://github.com/openT2T/translators/pull/75#issuecomment-242231119) [](ancestors = 242231119)

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.json
begillil: I'm not an expert on OCF, but I think that might be the way to go.

I have a different issue with that though. Is the expectation that once we use this to enumerate devices, the caller then initializes the translator with only this information, and then the translator then needs to do the full request to, for example Wink, for every single device that is enumerated?

To go from no devices to a list of populated devices then requires a new network request for every single device to get the same information that Wink provides from the '/users/me/wink_devices' endpoint, which seems fairly inefficient to me.

Is there a reason that we don't just return a list of populated OCF devices here?

---
In reply to: [76335569](https://github.com/openT2T/translators/pull/75#discussion_r76335569) [](ancestors = 76335569,76334693,76154047)

begillil:
That sounds reasonable.

---
In reply to: [242562378](https://github.com/openT2T/translators/pull/75#issuecomment-242562378) [](ancestors = 242562378,242231119)

org.opent2t.sample.hub.superpopular/org.opent2t.sample.hub.superpopular.json
arjun-msft: I'll look into the OCF bit.

On your other question, you're right it is a bit inefficient, the flow is the following: 

instantiate hub
get devices
using each item in the returned array, create a device translator
get details of each device

Now if you have a wink hub with 10 devices you'll end up with 11 calls to get the full details.  We need to support the alljoyn protocol and each of the translators need to run in that environment.  There we can't marshal objects back and forth only data, hence the need to instantiate each translator individually.

---
In reply to: [76336787](https://github.com/openT2T/translators/pull/75#discussion_r76336787) [](ancestors = 76336787,76335569,76334693,76154047)

begillil: In the alljoyn case, wouldn't you end up doing a call for each property of each device (at least how the translators are currently implemented)?

In that case, the number of calls seems to add up very quickly.

Is that not an expected scenario for the alljoyn protocol (enumerating devices and displaying the current state of each of them)?

---
In reply to: [76344264](https://github.com/openT2T/translators/pull/75#discussion_r76344264) [](ancestors = 76344264,76336787,76335569,76334693,76154047)

arjun-msft: Add the device type.

I agree we totally would end up doing a call per property and if we're going to the cloud that could potentially be hundreds of call and doesn't scale. AllJoyn is not something I have much experience in and we should let Jason comment on ways to optimize.  I'm not sure what will need to change to fully work in the AllJoyn world, but I know we can't pass objects around so we have this limitation on the design. I expect tweaks (hopefully) once we start working on enabling AllJoyn.

---
In reply to: [76345446](https://github.com/openT2T/translators/pull/75#discussion_r76345446) [](ancestors = 76345446,76344264,76336787,76335569,76334693,76154047)

begillil: Regarding support for both OCF and AllJoyn, couldn't we differentiate based on how we're calling into the translator?

We interact with OCF using getHubResURI, and we interact with AllJoyn using getDevices.

getHubResURI could return a list of OCF devices, while getDevices can return whatever AllJoyn would expect to get back, at least from the enumerate side, I suppose translator initialization would be impact in that case though.

This is probably fine for now, but I think we should at least think about something more efficient in the future.

---
In reply to: [76344264](https://github.com/openT2T/translators/pull/75#discussion_r76344264) [](ancestors = 76344264,76336787,76335569,76334693,76154047)

tjaffri:
:shipit: I vote we merge this... will merge in the morning unless somebody disagrees. Thanks, @arjun-msft !

PR #79, opened 2016-08-25
Fix baseresource dir casing, update readme

jasongin:
I noticed the "oic.baseresource" directory had an incorrect capital 'R', so I renamed the directory. The directory should match the case of the schema name which is always all-lowercase, not the filename that contains the schema, which may be mixed-case (or may even be different).

I added notes in the README about how to publish a translator package. It requires an extra manual step now due to a limitation of the script: https://github.com/openT2T/opent2t-cli/issues/7

I published the nest and wink translator packages using the steps described there. The npm "opent2t" account is an owner.


tjaffri:
:shipit:

PR #84, opened 2016-08-27
Dev

tjaffri:
- Bluetooth bulb, with unit tests for the flux helper. We should probably add some more unit tests for the translator itself.
- Wink bulb, with unit tests for on/off. I didn't implement dim/color but that should be easy to add on later. I did test the on/off functionality (that's identical to the BinarySwitch by the way)


arjun-msft:
@begillil Can you have a look at this?

org.opent2t.sample.lamp.superpopular/com.bluetoothlightbulb.fluxledlightbulb/js/thingTranslator.js
```diff
@@ -1,104 +1,208 @@
...
-                            });
-                        });
-                    });
-                } else {
-                    console.log('props.id is undefined');
-                }
+// in memory state of bulb
+var powered = false;
+var r = 0;
+var g = -0
```
begillil: -0? Also, a semicolon would be preferred.

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -1,110 +1,145 @@
...
-        },
-        method: 'PUT'
+    var powered = stateReader.get('powered');
+
+    return {
+        id: deviceSchema['object_type'] + '.' + deviceSchema['object_id'],
+        n: deviceSchema['name'],
+        rt: 'org.opent2t.sample.thermostat.superpopular',
+        power: { 'value': powered }
     };
```
begillil: We need to implement the rest of the lamp superpopular schema that is supported by Wink at some point (which I believe should be both dimming and color).

org.opent2t.sample.lamp.superpopular/com.bluetoothlightbulb.fluxledlightbulb/js/thingTranslator.js
tjaffri: Typo, fixing.

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
tjaffri: Yes, I ran out of time to do this... can you add that please? Please file an issue and follow up with a PR? I also didn't have a bulb connected to Wink that supported color/dimming so I wouldn't have been able to test it either even if I coded it.

begillil: Yeah, I'll look into doing this as part of a separate PR.

Not sure if I have a Wink bulb that supports color, but I should be able to find one at least that supports dimming.

---
In reply to: [76681179](https://github.com/openT2T/translators/pull/84#discussion_r76681179) [](ancestors = 76681179)

begillil:
:shipit:

Issue #86, opened 2016-08-30
Finish implementing Wink Lamp translator


begillil:
The Wink Lamp translator still needs to add support for dimming and color.

PR #87, opened 2016-08-31
Update Wink Lamp translator to add support for dim

begillil:
This is the dimming part of #86. I'm holding off on color for the moment, since the Lamp schema calls for RGB color, while Wink supports several different color formats, and so we will probably need to do some sort of conversion to support that.

### DIFF
--- a/org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
+++ b/org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
@@ -102,12 +102,12 @@ test.serial('PostLampResURI_Set_NameAndDim', t => {
 
             var value = {};
             value['n'] = "opent2t light";
-            value['dim'] = { 'dimmingSetting': 40 };
+            value['dim'] = { 'dimmingSetting': 43 };
 
             return OpenT2T.invokeMethodAsync(translator, 'org.opent2t.sample.lamp.superpopular', 'postLampResURI', [value]);
         }).then((response) => {
             t.is(response.n, "opent2t light");
-            t.is(response.dim.dimmingSetting, 40);
+            t.is(response.dim.dimmingSetting, 43);
 
             console.log('*** response: \n' + JSON.stringify(response, null, 2));
         });
--- a/org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
+++ b/org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
@@ -52,3 +52,64 @@ test.serial('Power', t => {
                 });
         });
 });
+
+// Get the entire Lamp schema object
+test.serial('GetLampResURI', t => {
+
+    return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config.Device)
+        .then(translator => {
+            // TEST: translator is valid
+            t.is(typeof translator, 'object') && t.truthy(translator);
+
+            return OpenT2T.invokeMethodAsync(translator, 'org.opent2t.sample.lamp.superpopular', 'getLampResURI', []);
+        }).then((response) => {
+            t.not(response.id, undefined);
+            t.is(response.rt, 'org.opent2t.sample.lamp.superpopular');
+            t.not(response.power, undefined);
+            t.not(response.dim, undefined);
+
+            console.log('*** response: \n' + JSON.stringify(response, null, 2));
+        });
+});
+
+
+// Get the entire Lamp schema object
+test.serial('PostLampResURI_Set_Power', t => {
+
+    return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config.Device)
+        .then(translator => {
+            // TEST: translator is valid
+            t.is(typeof translator, 'object') && t.truthy(translator);
+
+            var value = {};
+            value['power'] = { 'value': true };
+
+            return OpenT2T.invokeMethodAsync(translator, 'org.opent2t.sample.lamp.superpopular', 'postLampResURI', [value]);
+        }).then((response) => {
+            t.truthy(response.power.value);
+
+            console.log('*** response: \n' + JSON.stringify(response, null, 2));
+        });
+});
+
+// Set the name and dimming for the Lamp
+test.serial('PostLampResURI_Set_NameAndDim', t => {
+
+    return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config.Device)
+        .then(translator => {
+            // TEST: translator is valid
+            t.is(typeof translator, 'object') && t.truthy(translator);
+
+            var value = {};
+            value['n'] = "opent2t light";
+            value['dim'] = { 'dimmingSetting': 40 };
+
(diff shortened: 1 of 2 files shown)


org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -36,17 +36,29 @@ class StateReader {
     }
 }
 
+// Helper method to convert Wink's 0.0-1.0 brightness scale to a 0-100 scale
+function scaleBrightnessToDim(brightnessValue) {
+    return Math.floor(brightnessValue * 100);
+}
+
+// Helper method to convert a 0-100 scale to Wink's 0.0-1.0 brightness scale
+function scaleDimToBrightness(dimmingValue) {
```
begillil: If someone has a better idea for names for these two functions, I'm not super attached.

arjun-msft: I think the name is fine, a simpler name may be scaleBrightness

---
In reply to: [76908941](https://github.com/openT2T/translators/pull/87#discussion_r76908941) [](ancestors = 76908941)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -36,17 +36,29 @@ class StateReader {
...
+    return dimmingValue / 100;
```
arjun-msft: >100 [](start = 26, length = 3)

do you need a 100.0 here? I'm not sure how the rounding works here

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -36,17 +36,29 @@ class StateReader {
...
+
 // Helper method to convert the device schema to the translator schema.
 function deviceSchemaToTranslatorSchema(deviceSchema) {
     var stateReader = new StateReader(deviceSchema.desired_state, deviceSchema.last_reading);
 
     var powered = stateReader.get('powered');
+    var brightness = stateReader.get('brightness');
 
     return {
         id: deviceSchema['object_type'] + '.' + deviceSchema['object_id'],
```
arjun-msft: >id: deviceSchema['object_type'] + '.' + deviceSchema['object_id'] [](start = 8, length = 65)

why can't this simply return the object_id?

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
```diff
@@ -51,4 +51,65 @@ test.serial('Power', t => {
...
+// Set the name and dimming for the Lamp
+test.serial('PostLampResURI_Set_NameAndDim', t => {
+
+    return OpenT2T.createTranslatorAsync(translatorPath, 'thingTranslator', config.Device)
+        .then(translator => {
+            // TEST: translator is valid
+            t.is(typeof translator, 'object') && t.truthy(translator);
+
+            var value = {};
+            value['n'] = "opent2t light";
```
arjun-msft: >value['n'] = "opent2t light"; [](start = 12, length = 29)

totally didn't realize we supported this, very cool!

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
```diff
@@ -51,4 +51,65 @@ test.serial('Power', t => {
...
+            value['dim'] = { 'dimmingSetting': 40 };
```
arjun-msft: >40 [](start = 47, length = 2)

i wonder if a value like 43 may be a better test (for validating rounding)

arjun-msft:
:clock1:

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
begillil: I like that idea better. I'll change it.

---
In reply to: [76938899](https://github.com/openT2T/translators/pull/87#discussion_r76938899) [](ancestors = 76938899,76908941)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
begillil: That was my first thought as well, but I tried it out and it's fine as is. (e.g. 43 / 100 = 0.43; C-style integer/floating point rules don't apply in JavaScript, I suppose)

---
In reply to: [76938955](https://github.com/openT2T/translators/pull/87#discussion_r76938955) [](ancestors = 76938955)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
begillil: It was somewhat arbitrary, but for Wink, the minimum set of information to interact with a specific device is the object type and the object ID, so that's what I went with when implementing the thermostat. For the Lamp, I believe the same style carried over from that decision.

I think it would probably be good to have a deeper discussion about how we should treat this ID field.

With Wink in particular, there is a problem with trying to use just the object_id to identify a device over time, since, according to Wink's documentation:

"It is possible for the API to re-assign identifiers to resources to rebalance keys; in this case, your resource will still exist but it (and all references to it) will be updated to the new identifier. Your application should be able to handle this case."

Wink exposes a uuid field that should uniquely identify a device, regardless of its object_id, however, you cannot actually interact with the device without the object_id.

I think there is value in exposing the uuid field, since without it, I think it's impossible for an application using openT2T to handle the case where the object_id is re-assigned, so I would prefer that we use that here, but then we don't have the object_id in the device schema at all, so I don't know if that breaks any scenarios.

---
In reply to: [76939288](https://github.com/openT2T/translators/pull/87#discussion_r76939288) [](ancestors = 76939288)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/tests/test.js
begillil: It at least works for lights. Despite Wink's documentation showing that the 'name' field is writable on devices in general, when I tried this out on a Nest thermostat attached to Wink, the REST calls were failing. I noticed that the name isn't editable in their app either, so I don't know if that is just a specific limitation for the Nest thermostat or thermostats in general.

---
In reply to: [76939487](https://github.com/openT2T/translators/pull/87#discussion_r76939487) [](ancestors = 76939487)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
arjun-msft: Makes sense. I logged issue 89 to track this.

---
In reply to: [77048026](https://github.com/openT2T/translators/pull/87#discussion_r77048026) [](ancestors = 77048026,76939288)

PR #93, opened 2016-09-13
Updating eslint config based on changes to the opent2t-config package.

ChuckFerring:


arjun-msft:
Looks good.  Verified the test.

PR #94, opened 2016-09-13
Updating wink lightbulb to depend on the wink hub

arjun-msft:


arjun-msft:
Another PR has this change.

PR #99, opened 2016-09-16
Adding realtime eventing to Wink hub, bulb, and thermostat

danielrischar:
This adds real-time notifications to the wink devices.  The hub offers a way for subscribe a callback to all changes on the hub/provider.  The callback will receive a JSON object representing the translated device schema in it's new state.

### DIFF
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
@@ -6,54 +6,77 @@
 
 "use strict";
 var request = require('request-promise');
+var PubNub = require('pubnub');
+var EventEmitter = require('events');
 
 /**
 * This translator class implements the "Hub" interface.
 */
-class Translator {
+class Translator extends EventEmitter {
     constructor(accessToken) {
+        super();
+
         this._accessToken = accessToken;
 
         this._baseUrl = "https://api.wink.com";
+        this._userPath = '/users/me';
         this._devicesPath = '/users/me/wink_devices';
 
         this._name = "Wink Hub"; // TODO: Can be pulled from OpenT2T global constants. This information is not available, at least, on wink hub.
+
+        this._eventTranslators = {};
     }
 
     /**
      * Get the list of devices discovered through the hub.
      */
     getHubResURI() {
-        return this._makeRequest(this._devicesPath, 'GET').then((response) => {
+        return this._makeRequest(this._userPath, 'GET').then((response) => {
+            // Subscribe to the user feed for device discovery
+            // TODO: This doesn't work, at least for non-wink devices like Nest thermostats
+            // May need to contact Wink.
+            this._subscribe(response.data.subscription.pubnub);
 
-            var toReturn = {};
-            var devices = response.data;
+            return this._makeRequest(this._devicesPath, 'GET').then((response) => {
+                this._pubNubChannels = [];
+                this._pubNubSubscribeKey;
 
-            var filteredDevices = [];
-            devices.forEach((winkDevice) => {
-                // get the opent2t schema and translator for the wink device
-                var opent2tInfo = this._getOpent2tInfo(winkDevice);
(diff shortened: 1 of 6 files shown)


org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -89,6 +90,88 @@ class Translator {
...
+                    hub: this,
+                    deviceInfo: { id: 0 } // Device identifier isn't needed as the translator will not be used to
+                                          // communicate with an API endpoint.
+                };
+
+                // TODO: There must be a better way of doing this that escapes me at the moment
+                // We're not truly creating a translator, I just want access to the 'deviceSchemaToTranslatorSchema'
+                // function
+                var translatorName = opent2tinfo.translator.replace(/opent2t-translator-/gi, "").replace(/-/g, ".");
+                var translatorPath = require('path').join(__dirname, '../../..', opent2tinfo.schema, translatorName, "js", "thingTranslator.js");
```
arjun-msft: >var translatorPath = require('path').join(__dirname, '../../..', opent2tinfo.schema, translatorName, "js", "thingTranslator.js"); [](start = 16, length = 129)

relative paths may get you in to trouble, use OpenT2T.createTranslatorAsync to do this.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -89,6 +90,88 @@ class Translator {
         return this._makeRequest(requestPath, 'PUT', putPayloadString);
     }
 
+    /**
+     * Subscribes to realtime notifications for this provider.
+     * 
+     * callback is the function that will be called with the JSON ocf representation of the changed device
+     */
+    subscribe(callback) {
```
arjun-msft: >subscribe [](start = 4, length = 9)

where's the schema change to go along with this?

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -89,6 +90,88 @@ class Translator {
         return this._makeRequest(requestPath, 'PUT', putPayloadString);
     }
 
+    /**
+     * Subscribes to realtime notifications for this provider.
```
arjun-msft: We should also have an unsubscribe

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -149,6 +112,43 @@ class Translator {
                 return response.power.value;
             });
     }
+
+    // Helper method to convert the device schema to the translator schema.
+    deviceSchemaToTranslatorSchema(deviceSchema) {
```
arjun-msft: >deviceSchemaToTranslatorSchema [](start = 4, length = 30)

add static to both of these functions so you don't need an instance to reference. We have an example of this in the OpenT2T library also (it's in ts but when you compiles shows you the js syntax)
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static)

arjun-msft:
:clock1:

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
arjun-msft: in the hub raml file, add another method like so:

/SubscribeHubResURI, make it a post, the request body should have a blob (it'll be a callback but for raml make it a blob), and the response should be the same as the get. the response ties in with the what the user will see on the notification.  it's a convention we are enabling through our javascript implementation

---
In reply to: [79207243](https://github.com/openT2T/translators/pull/99#discussion_r79207243) [](ancestors = 79207243)

PR #102, opened 2016-09-20
Updating package.json for publish

arjun-msft:


arjun-msft:
fyi these have been published and validated.

org.opent2t.sample.thermostat.superpopular/org.opent2t.sample.thermostat.superpopular.json
```diff
@@ -91,6 +91,18 @@
...
+    },
+    {
+      "$ref": "oic.r.clock.json#/definitions/oic.r.clock"
+    },
+    {
+      "$ref": "oic.r.temperature.json#/definitions/oic.r.temperature"
+    },
+    {
+      "$ref": "oic.r.humidity.json#/definitions/oic.r.humidity"
+    },
```
begillil: Is there a specific reason we're adding these here?

arjun-msft: I thought if they are referenced above they need to be here. is that the wrong assumption?

---
In reply to: [79676179](https://github.com/openT2T/translators/pull/102#discussion_r79676179) [](ancestors = 79676179)

begillil: That isn't my understanding. I believe this is where we actually describe all of the things that we want this schema to represent. Above we describe a 'definition' for org.opent2t.sample.thermostat.superpopular, and we reference that here, along with oic.core and oic.r.baseresource (which, looking at it, I'm not sure makes sense here either).

For example, by including oic.r.temperature here, I believe that's saying we want to return an object that looks like:

{
  "id": "<id>",
  "rt": "org.opent2t.sample.thermostat.superpopular",
  "ambientTemperature": { <temperature> }
  . . . (the rest of the Thermostat properties)
  "temperature": 52
  "units": "F"
}

Which isn't what we're trying to describe.

---
In reply to: [79681760](https://github.com/openT2T/translators/pull/102#discussion_r79681760) [](ancestors = 79681760,79676179)

begillil: See the Airflow Control for an example:

[http://oneiota.org/revisions/1485](http://oneiota.org/revisions/1485)

---
In reply to: [79683703](https://github.com/openT2T/translators/pull/102#discussion_r79683703) [](ancestors = 79683703,79681760,79676179)

arjun-msft: removing these.

---
In reply to: [79685801](https://github.com/openT2T/translators/pull/102#discussion_r79685801) [](ancestors = 79685801,79683703,79681760,79676179)

PR #105, opened 2016-09-28
Fixing invalid RAML by adding subscription schema

danielrischar:
Invalid RAML causes OpenT2T to choke when calling getSchemaAsync(), need to have subscription schemas.


danielrischar:
I'm also looking to add a test for this so that we can verify future RAML changes easily.

avneet723:
Let's make sure that we use the test you are writing and onboard it as part of the ES lint gate for new translators.

PR #111, opened 2016-10-19
Add SmartThings hub, lamp, and binary switch translators.

m-chau:


arjun-msft:
Hey Mary, Avneet verified this works as well, we are going to merge without doing a full CR because we are moving to the new three tiered resource model which should come to Wink by end of week which will require updates to this. So postponing full code review until then.

PR #116, opened 2016-10-25
Switching CI tasks to all use gulp to enable running tasks within sub…

ChuckFerring:
…directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,30 @@ Follow our getting started guide at http://www.opentranslatorstothings.org. Note
 Here is some background reading for those who are curious:
 1. Node package name requirements/rules: https://docs.npmjs.com/files/package.json.
 2. Issue #50 includes a discussion and some context behind this naming guidance.
+
+## Run Integration Tests
+
+1. Install gulp globally.
+```bash
+npm install -g gulp
+```
+
+2. Install dependencies.
+```bash
+npm install
+```
+
+3. Run integration tests.
+```bash
+gulp ci-checks
+```
+
+Notes:
+1. Other gulp tasks can be run as well, see gulpfile.js for available tasks.
+2. By default all files under the translators repo will be tested.  use the --cwd option to only test files under a specified directory:
+```bash
+gulp --cwd .\org.opent2t.sample.windowshade.superpopular\ ci-checks
+```
  
 ## Create a Pull Request
 Made any changes we should consider? Send us a pull request! Check out [this article](https://help.github.com/articles/creating-a-pull-request/)
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,7 +6,11 @@ var eslint = require('gulp-eslint');
 var gutil = require('gulp-util');
 var through = require('through2');
 var path = require('path');
+const NO_MODULES = '!**/node_modules/**';
 
+// The namingValidator validates that the name of the translator is consistent
+// between the translator folder and the package.json file.  It also validates
+// that the name of the translator is correctly prepended in the package.json file.
 var namingValidator = function() {
   const PLUGIN_NAME = 'Naming Validator';
   const NAME_PREFIX = 'opent2t-translator-';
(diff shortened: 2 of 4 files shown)


gulpfile.js
```diff
@@ -1,7 +1,78 @@
+"use strict";
+
 var gulp = require('gulp');
 var xmlValidator = require('gulp-xml-validator');
+var eslint = require('gulp-eslint');
+var gutil = require('gulp-util');
+var through = require('through2');
+var path = require('path');
+
+var namingValidator = function() {
```
arjun-msft: add some comments, what does this function do?

gulpfile.js
```diff
@@ -1,7 +1,78 @@
...
+     }
+     callback();
+  }
+
+  return through.obj(transformFunction, errorOutput);
+}
+
+gulp.task('validate-naming', function () {
+  return gulp.src(['**/js/package.json', '!**/node_modules/**', '!**/Helpers/**'])
+    .pipe(namingValidator())
```
arjun-msft: do you need .format and .failAfterError here also?

package.json
```diff
@@ -1,18 +1,16 @@
 {
```
arjun-msft: looks like it does, the pack-translator *will* overwrite this file. sounds like we'll need to update pack-translator

---
In reply to: [85858482](https://github.com/openT2T/translators/pull/116#discussion_r85858482) [](ancestors = 85858482)

README.md
```diff
@@ -30,6 +30,30 @@ Follow our getting started guide at http://www.opentranslatorstothings.org. Note
 Here is some background reading for those who are curious:
```
shalinijoshi19: Sorry if I missed this but what is gulp and what value is it adding again?

gulpfile.js
```diff
@@ -1,7 +1,78 @@
...
 gulp.task('xml-lint', function () {
-  return gulp.src('**/*.xml')
+  return gulp.src(['**/*.xml', '!**/node_modules/**'])
     .pipe(xmlValidator())
-});
\ No newline at end of file
+});
+
+gulp.task('js-lint', () => { 
+    return gulp.src(['**/*.js','!**/node_modules/**'])
```
shalinijoshi19: Instead of passing the regex over and over can we assign it to a variable instead?

package.json
```diff
@@ -1,18 +1,16 @@
...
     "eslint-config-opent2t": "^1.0.2",
     "eslint-plugin-json": "^1.2.0",
     "eslint-plugin-raml": "^0.2.6",
     "gulp": "^3.9.1",
-    "gulp-xml-validator": "^0.1.2"
+    "gulp-eslint": "^3.0.1",
+    "gulp-util": "^3.0.7",
+    "gulp-xml-validator": "^0.1.2",
+    "through2": "^2.0.1"
   }
```
shalinijoshi19: Curious -- how did you arrive at these dependencies again?

gulpfile.js
ChuckFerring: format and failafter are part of the eslint gulp plugin which isn't used for this task.

---
In reply to: [85858380](https://github.com/openT2T/translators/pull/116#discussion_r85858380) [](ancestors = 85858380)

README.md
ChuckFerring: Gulp is a very commonly used build system/task runner.  The main benefits are that it has lot's of extensions for things like eslint, ava, xml validation, etc and that it is an easy framework for running file glom based tasks.  Originally we were doing most of the tasks as simple npm scripts defined in the root package.json but this didn't work when trying to run the validation scripts in subdirectories that had their own package.json files because npm expected the scripts to be defined in the nearest ancestor directory with a package.json file.

---
In reply to: [85870131](https://github.com/openT2T/translators/pull/116#discussion_r85870131) [](ancestors = 85870131)

package.json
ChuckFerring: gulp-eslint is needed because the eslint tasks are now run through gulp.
gulp-util and through2 are needed for the specialized name validation task that was added.

---
In reply to: [85870412](https://github.com/openT2T/translators/pull/116#discussion_r85870412) [](ancestors = 85870412)

arjun-msft:
:shipit:

PR #119, opened 2016-11-11
Adding additional CI verifiers to validate manifest files and raml sc…

ChuckFerring:
…hema compliance.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,56 +1,26 @@
 "use strict";
 
+const NO_MODULES = '!**/node_modules/**';
 var gulp = require('gulp');
 var xmlValidator = require('gulp-xml-validator');
 var eslint = require('gulp-eslint');
-var gutil = require('gulp-util');
-var through = require('through2');
-var path = require('path');
-const NO_MODULES = '!**/node_modules/**';
+var InterfaceValidator = require('./verifiers/interfaceValidator');
+var NamingfaceValidator = require('./verifiers/namingValidator');
+var ManifestValidator = require('./verifiers/manifestValidator');
 
-// The namingValidator validates that the name of the translator is consistent
-// between the translator folder and the package.json file.  It also validates
-// that the name of the translator is correctly prepended in the package.json file.
-var namingValidator = function() {
-  const PLUGIN_NAME = 'Naming Validator';
-  const NAME_PREFIX = 'opent2t-translator-';
-  var errorOutputList = [];
-
-  function transformFunction (file, encoding, callback) {
-    if (file.isNull()) {
-      return callback(null, file);
-    }
-
-    if (file.isStream()) {
-      return callback(new gutil.PluginError(PLUGIN_NAME, 'Streaming not supported'));
-    }
-
-    var metadata = JSON.parse(file.contents.toString());
-    var folderName = path.basename(path.dirname(path.dirname(file.path)));
-    var expectedName = NAME_PREFIX + folderName.replace(/\./g, '-');
-
-    if(metadata.name !== expectedName) {
-        errorOutputList.push(file.path + '\n' + gutil.colors.red("Error: ") + 'Incorrect translator name. Expected: ' + expectedName);
-    }
-
-    return callback(null, file);
-  }
-
-  function errorOutput(callback) {
-    if (errorOutputList.length > 0) {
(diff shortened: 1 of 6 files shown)


verifiers/interfaceValidator.js
```diff
@@ -0,0 +1,85 @@
...
+
+  while(prevDir !== currentDir) {
+    var ramlFiles = glob.sync(currentDir + '/*.raml');
+
+    if(ramlFiles.length > 0) {
+      return ramlFiles[0];
+    }
+
+    prevDir = currentDir;
+    currentDir = path.join(currentDir, '..');
```
arjun-msft: does this fail when it hits the root or something? how does this end?

verifiers/interfaceValidator.js
```diff
@@ -0,0 +1,85 @@
+"use strict";
```
arjun-msft: we should add unit tests to each of these

verifiers/manifestValidator.js
```diff
@@ -0,0 +1,70 @@
...
+                match = regex.exec(urlTemplate);
+                if (match && urlTokens.indexOf(match[1]) === -1 ) {
+                    urlTokens.push(match[1]);
+                }
+            } while (match);
+        }
+    }
+
+    urlTokens.forEach((value) => {
+        var valueNode = xpath.select("onboardflow[@name='getDeveloperInput']/flow[arg[@name='" + value + "']]", onboardingNode, true);
```
arjun-msft: they could also be defined in getUserInput

verifiers/manifestValidator.js
```diff
@@ -0,0 +1,70 @@
...
+// If a manifest has onboarding defined, verify the onboarding id
+// matches a dependency in the package.json file.
+function validateNaming(onboardingNode, log, filepath) {
+    const NAME_PREFIX = 'opent2t-onboarding-';
+    var id = xpath.select1("@id", onboardingNode);
+    var expectedDependencyName = NAME_PREFIX + id.value.replace(/\./g, '-');
+    var packagePath = path.join(path.dirname(filepath), 'package.json');
+    var metadata = JSON.parse(fs.readFileSync(packagePath).toString());
+
+    if(!metadata.dependencies.hasOwnProperty(expectedDependencyName)) {
```
arjun-msft: >hasOwnProperty [](start = 30, length = 14)

what happens if there is a version specified? does this handle that case? opent2t-onboarding-org-opent2t-onboarding-winkhub@1.2

gulpfile.js
```diff
@@ -1,78 +1,48 @@
 "use strict";
 
+const NO_MODULES = '!**/node_modules/**';
```
arjun-msft: >NO_MODULES [](start = 6, length = 10)

nit/question: why NO instead of NODE

verifiers/interfaceValidator.js
arjun-msft: scenario is what if there isn't any raml files in any parent folder

---
In reply to: [87661150](https://github.com/openT2T/translators/pull/119#discussion_r87661150) [](ancestors = 87661150)

arjun-msft:
:clock1:

verifiers/interfaceValidator.js
ChuckFerring: Yes, this will stop looking after it reaches the root directory.  If no raml file is found by then it assumes there are no schema methods to look for.

---
In reply to: [87662212](https://github.com/openT2T/translators/pull/119#discussion_r87662212) [](ancestors = 87662212,87661150)

gulpfile.js
ChuckFerring: This is meant to exclude anything found in node_modules folders so the NO is meant to meant "don't look in modules folders", it's not meant as an abbreviation of NODE.

---
In reply to: [87662095](https://github.com/openT2T/translators/pull/119#discussion_r87662095) [](ancestors = 87662095)

verifiers/manifestValidator.js
arjun-msft: ignore this, got my references mixed up

---
In reply to: [87661890](https://github.com/openT2T/translators/pull/119#discussion_r87661890) [](ancestors = 87661890)

verifiers/interfaceValidator.js
arjun-msft: discussed offline, created a deliverable

---
In reply to: [87683181](https://github.com/openT2T/translators/pull/119#discussion_r87683181) [](ancestors = 87683181,87661373)

Issue #74, opened 2016-08-24
Update hub translator to do automatic refresh of token if applicable


arjun-msft:
In situations where the hub can do a refresh of the access token it should do that, such as in the case of Wink.

arjun-msft:
This is no longer applicable, specific errors will be raised and refresh will be called to do a refresh.

PR #126, opened 2016-11-21
Implementing refreshAuthToken API in hub translator

shalinijoshi19:


shalinijoshi19:
I accidentally messed up a local squash and it looks like Ben's description got applied to my change anyways :( -- I am not sure how to 'unsquash'

opent2t.p.hub/com.wink.hub/js/thingTranslator.js
```diff
@@ -284,14 +322,19 @@ class Translator {
     /**
      * Internal helper method which makes the actual request to the wink service
      */
-    _makeRequest(path, method, content) {
+    _makeRequest(path, method, content, includeBearerHeader) {
+       
+        var addBearerHeader = (typeof includeBearerHeader !== 'undefined') ?  includeBearerHeader : true;
```
shalinijoshi19: yeah that's what the initial implementation was based on but turns out that the version of js we are using (am not sure how I can figure that out yet) isn't ECMA6.. we'll need to follow up offline on why we aren't using the latest and greatest yet.

arjun-msft: This seems a bit complicated, why not just check if includeBearerHeader is true below?

opent2t.p.hub/com.wink.hub/js/thingTranslator.js
```diff
@@ -108,6 +110,42 @@ class Translator {
         // Make the async request
         return this._makeRequest(requestPath, 'PUT', putPayloadString);
     }
+
+    /**
+     * Refreshes the OAuth token for the hub by sending a refresh POST to the wink provider
+     */
+    refreshAuthToken(authInfo)
+    {
+        console.log("In refreshAuthToken for WINK hub..");
```
arjun-msft: I don't see this logging in other methods, should this be removed?

arjun-msft:
:shipit:

opent2t.p.hub/com.wink.hub/js/thingTranslator.js
shalinijoshi19: I can remove it - we are inconsistent though -- onboarding for eg we do have a similar log statement. do we have a story yet for replacing console.logs with 'wrapperLibraryForTelemetry.log' ?

opent2t.p.hub/com.wink.hub/js/thingTranslator.js
```diff
@@ -361,7 +369,6 @@ class Translator {
             .catch(function (err) {
                 console.log("Request failed to: " + options.method + " - " + options.url);
                 console.log("Error            : " + err.statusCode + " - " + err.response.statusMessage);
-                // todo auto refresh in specific cases, issue 74
```
shalinijoshi19: @arjun-msft , I couldn't find issue 74 - where is this referenced; Since the 'auto' part of refresh would be within the cloud federated service, I guess the comment doesn't quiet apply? What might still apply and what I am not quite sure yet is if any specific error code is returned by WINK that we need to make the caller (Federated service in this case) Aware of to allow them to call into the refreshAuthToken() API.

arjun-msft: [https://github.com/openT2T/translators/issues/74](https://github.com/openT2T/translators/issues/74)

we can't do auto-refresh, I'm going to close the issue.  We need an explicit call to do refresh.

---
In reply to: [88992135](https://github.com/openT2T/translators/pull/126#discussion_r88992135) [](ancestors = 88992135)

opent2t.p.hub/com.wink.hub/js/thingTranslator.js
shalinijoshi19: it appears that this was the defacto std way of setting a default value for function parameters ie default parameters-- so I didn't want to have to change all existing calls to have it be true so if not specified this way I can default it to true


---
In reply to: [88971888](https://github.com/openT2T/translators/pull/126#discussion_r88971888) [](ancestors = 88971888)

shalinijoshi19: i see your point -- update coming up

---
In reply to: [89002893](https://github.com/openT2T/translators/pull/126#discussion_r89002893) [](ancestors = 89002893,88971888)

shalinijoshi19: since this is a stop gap at the moment ntil we figure out ohw to upgrade to latest JS that supports default params, keeping it as is

---
In reply to: [89005342](https://github.com/openT2T/translators/pull/126#discussion_r89005342) [](ancestors = 89005342,89002893,88971888)

arjun-msft: That's fine.

---
In reply to: [89006587](https://github.com/openT2T/translators/pull/126#discussion_r89006587) [](ancestors = 89006587,89005342,89002893,88971888)

PR #128, opened 2016-11-22
Subscription update

danielrischar:
Moving translation method into translator class and adding smarthings platform(s) translation.


arjun-msft:
Did you verify everything with unit tests and/or cli?

arjun-msft:
:shipit:

PR #130, opened 2016-11-28
Renaming translators to org.opent2t.sample.*.superpopular

danielrischar:
Bulk renaming of the new translator models into the old naming style.


danielrischar:
I've verified this rename by creating fake npm packages using pack-translator.js and npm pack, and using those packages with both the opent2t-cli and cloud tests.

org.opent2t.sample.binaryswitch.superpopular/org.opent2t.sample.binaryswitch.superpopular.xml
```diff
@@ -1,11 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
arjun-msft: why is this being deleted/updated?

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "opent2t-translator-com-smartthings-binaryswitch",
-  "version": "1.0.0",
+  "version": "0.2.0",
```
arjun-msft: not sure if versions can go down/backward. Avneet mentioned something about this.

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/tests/test.js
```diff
@@ -9,12 +8,12 @@ console.log(JSON.stringify(config, null, 2));
 var translatorPath = require('path').join(__dirname, '..');
 var hubPath = require('path').join(__dirname, '../../../../org.opent2t.sample.hub.superpopular/com.smartthings.hub/js');
 var translator = undefined;
+var controllId = undefined;
```
arjun-msft: spelling, there are two Ls

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -154,51 +154,49 @@ void updateConnectionId() {
     log.debug "Updated ConnectionID to ${state.connectionId}"
 }
 
-//Endpoints functions; return device data in json format
```
arjun-msft: did you update this file or is this part of a merge?

org.opent2t.sample.lamp.superpopular/org.opent2t.sample.lamp.superpopular.xml
```diff
@@ -1,20 +0,0 @@
...
-    <property name="dim" type="d" access="readwrite">
-      <description>Dimming status of the light.</description>
-    </property>
-
-    <property name="color" type="ai" access="readwrite">
-      <description>RGB color emitted by the light, e.g. [255, 255, 255]</description>
-    </property>
-
-  </interface>
-</node>
```
arjun-msft: again, don't think we should delete these files

arjun-msft:
:clock1:

arjun-msft:
A few tiny things, but mostly good.  Two open question, 1. deleting xml schema files, 2. going backward in version numbers.

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
danielrischar: We never published v1.0.0 of this translator, the latest on NPM is 0.0.1 so we can move to 0.2.0 to represent the new resource model.

---
In reply to: [90066887](https://github.com/openT2T/translators/pull/130#discussion_r90066887) [](ancestors = 90066887)

avneet723: Current NPM packages for SmartThings are at 0.0.1, so updating to 0.2.0 for the new resource model should be fine. I propose that version is updated via a PR before publish to NPM so our GitHub main has the latest version at all times.

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
danielrischar: All I am doing in this is moving the new changes into the old folders.  Git views this as a bunch of edits rather than just replacing the files of the same name.  Any changes are the ones we already have in opent2t.p.* folders.

---
In reply to: [90067590](https://github.com/openT2T/translators/pull/130#discussion_r90067590) [](ancestors = 90067590)

danielrischar:
The version numbers are good (we never published the higher ones).

Any edits to files are a result of git thinking I edited them rather than copied the new versions (from opent2t.p.* folders) over the old ones (from org.opent2t.sample.*.superpopular).  Git views this as an edit because the files have the same name, though more appropriately it is a delete, a move, and a rename.

I can put back the XML, but we did not use these in the new opentt2t.p.* naming so I don't know what the value is to having them.

---
In reply to: [263649123](https://github.com/openT2T/translators/pull/130#issuecomment-263649123) [](ancestors = 263649123)

PR #135, opened 2016-11-30
Updated unit tests and device tests for some translators to leverage …

ChuckFerring:
…shared tests and mock devices.

The goal is to create a reuse pattern for sharing tests among devices of the same type.  Tests are also shared between unit tests that run against mocks and device tests.

### DIFF
--- a/org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
+++ b/org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
@@ -10,7 +10,8 @@
   "main": "thingTranslator.js",
   "license": "MIT",
   "scripts": {
-    "test": "ava"
+    "test": "ava tests/test.js",
+    "unittest": "ava tests/unitTests.js"
   },
   "ava": {
     "files": [
@@ -24,6 +25,8 @@
   "devDependencies": {
     "optimist": "*",
     "ava": "^0.15.2",
-    "es6-sleep": "*"
+    "es6-sleep": "*",
+    "opent2t-device-smartthingshub": "file:../../../org.opent2t.sample.hub.superpopular/com.smartthings.hub/test",
+    "opent2t-device-binaryswitch": "file:../../test"
   }
 }
--- a/org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/tests/test.js
+++ b/org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/tests/test.js
@@ -1,13 +1,10 @@
 var test = require('ava');
 var OpenT2T = require('opent2t').OpenT2T;
-var config = require('./testConfig');
-
-console.log("Config:");
-console.log(JSON.stringify(config, null, 2));
-
 var translatorPath = require('path').join(__dirname, '..');
+var runBinarySwitchTests = require('opent2t-device-binaryswitch/binarySwitchTests');
+var config = require('./testConfig');
 var hubPath = require('path').join(__dirname, '../../../../org.opent2t.sample.hub.superpopular/com.wink.hub/js');
-var translator = undefined;
+var deviceId = "F85B0738-6EC0-4A8B-A95A-503B6F2CA0D8";
 
 function getBinarySwitch(devices) {
     for (var i = 0; i < devices.length; i++) {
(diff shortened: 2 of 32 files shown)


org.opent2t.sample.binaryswitch.superpopular/test/binarySwitchTests.js
```diff
@@ -0,0 +1,71 @@
...
+function runBinarySwitchTests(createTranslator, deviceId, test, setTestData) {
+
+    function setData(t) {
+        if(setTestData) {
+            setTestData(t.title, t);
+        }
+    }
+
+    test.serial('Valid Binary Switch Translator', t => {
+        return createTranslator().then(translator => {
```
arjun-msft: >createTranslator [](start = 15, length = 16)

can we do this once at the beginning of the test rather than each time?

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
```diff
@@ -24,6 +25,8 @@
   "devDependencies": {
     "optimist": "*",
     "ava": "^0.15.2",
-    "es6-sleep": "*"
+    "es6-sleep": "*",
+    "opent2t-device-smartthingshub": "file:../../../org.opent2t.sample.hub.superpopular/com.smartthings.hub/test",
+    "opent2t-device-binaryswitch": "file:../../test"
```
arjun-msft: Is this getting the mock devices?

---
In reply to: [90383696](https://github.com/openT2T/translators/pull/135#discussion_r90383696) [](ancestors = 90383696)

org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/tests/test.js
```diff
@@ -1,13 +1,10 @@
 var test = require('ava');
 var OpenT2T = require('opent2t').OpenT2T;
-var config = require('./testConfig');
-
-console.log("Config:");
-console.log(JSON.stringify(config, null, 2));
-
 var translatorPath = require('path').join(__dirname, '..');
+var runBinarySwitchTests = require('opent2t-device-binaryswitch/binarySwitchTests');
+var config = require('./testConfig');
```
arjun-msft: >testConfig [](start = 24, length = 10)

anyway to document the structure/shape of this file?

arjun-msft: and everyone of them in any folder

---
In reply to: [90384226](https://github.com/openT2T/translators/pull/135#discussion_r90384226) [](ancestors = 90384226)

arjun-msft:
:shipit:

arjun-msft:
I think this looks good, I recommend @danielrischar or @shalinijoshi19 double checking this because of the size of the change.

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
ChuckFerring: Yes, it's getting the mock hub and the mock binary switch.  the "[file:...](file:...)" syntax tells npm to look for the package locally rather than trying to obtain it from a server.  This allows us to package the mock devices and require them like any other node package without having to publish them.

---
In reply to: [90383696](https://github.com/openT2T/translators/pull/135#discussion_r90383696) [](ancestors = 90383696)

shalinijoshi19: that's actually cool that you can use this syntax to tell npm to pull files locally;  Is there a way to tell it to pull it locally + from teh server (fallback to local incase server fails) or something?

shalinijoshi19:
took a quick look - as long as the tests that you fixed are succeeding and the steps to run them is doc'd am all in for this :) thanks a ton!

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/package.json
shalinijoshi19: hmmm does this mean this will always need the test files even for say the CLI? how does this work ? should this be a test dependency and not a dev dependency?

---
In reply to: [90574294](https://github.com/openT2T/translators/pull/135#discussion_r90574294) [](ancestors = 90574294)

shalinijoshi19:
:shipit:

org.opent2t.sample.binaryswitch.superpopular/com.wink.binaryswitch/js/tests/test.js
```diff
@@ -1,13 +1,10 @@
...
-
-console.log("Config:");
-console.log(JSON.stringify(config, null, 2));
-
 var translatorPath = require('path').join(__dirname, '..');
+var runBinarySwitchTests = require('opent2t-device-binaryswitch/binarySwitchTests');
+var config = require('./testConfig');
 var hubPath = require('path').join(__dirname, '../../../../org.opent2t.sample.hub.superpopular/com.wink.hub/js');
-var translator = undefined;
+var deviceId = "F85B0738-6EC0-4A8B-A95A-503B6F2CA0D8";
```
m-chau: >F85B0738-6EC0-4A8B-A95A-503B6F2CA0D8 [](start = 16, length = 36)

Is there a reason why we are hard coding the deviceID instead of getting it from platfromInfo during the translator setup?

ChuckFerring: Two reasons for this.
1.  That ID is actually hard-coded in the thingTranslator.
2.  It's the same ID that was already being passed in (look at the old implementation of GetPower and SetPower tests).  I just pulled it out into a var.

---
In reply to: [90578036](https://github.com/openT2T/translators/pull/135#discussion_r90578036) [](ancestors = 90578036)

PR #138, opened 2016-12-02
Updated interface validator to construct raml method names correctly.

ChuckFerring:


verifiers/interfaceValidator.js
```diff
@@ -44,17 +44,52 @@ function getRamlFilePath(filePath) {
...
-    raml.loadApiSync(ramlPath).resources().forEach(resource => {
-      resource.methods().forEach(method => {
-        ramlMethods.push(method.method() + resource.completeRelativeUri().substring(1));
-      });
+    var resources = [];
+    raml.loadApiSync(ramlPath).resources().forEach(r => {
+      resources.push(r);
     });
+
+    while(resources.length > 0) {
```
arjun-msft: I'd recommend adding comments for each section with examples of what is being looked at and expected

verifiers/interfaceValidator.js
```diff
@@ -44,17 +44,52 @@ function getRamlFilePath(filePath) {
...
+
+      current.methods().forEach(method => {
+        addMethod(method.method() + suffix);
+      });
+
+      current.resources().forEach(child => {
+        resources.push(child);
+      });
+    }
   }
```
avneet723: How do we validate that any GET or POST calls in the thingTranslator.js are also defined in the RAML? For instance, refresh mechanism was added to Wink but it got left out from RAML. We will fix this for now but it would be good to catch any such future instances.

Issue #142, opened 2016-12-21
Translators' properties do not persist when there are multiple translators of the same device type.


m-chau:
Currently when we enumerate platforms of the same provider, we create a translator for each platform, then use that translator to invoke the ‘get’ method to get the platform details. When I have two outlet (binary switch) devices and call the hub translator to enumerate them, the control ID of the first outlet translator gets overwritten by that of the second outlet device.

Below is the log I get by adding some debug output. You may see the source for SmartThings hub at my repository . 
```
SmartThings Binary Switch initializing...
    Initialize translator for control ID: 529fa2ef-668b-4e84-a101-c47968997d8a
SmartThings Binary Switch initializing...Done                                   => translator created for outlet 1

SmartThings Binary Switch initializing...
    Initialize translator for control ID: e7a7c755-647e-4b69-8853-d66af047be4c
SmartThings Binary Switch initializing...Done                                   => translator created for outlet 2

* HUB::Invoking GET method for control ID: 529fa2ef-668b-4e84-a101-c47968997d8a => before line 110 in SmartThings hub: OpenT2T.invokeMethodAsync(...) to get device details for outlet 1
** Translator::get for control ID: e7a7c755-647e-4b69-8853-d66af047be4c         => In the “get(expand, payload)” method of outlet 1’s translator

* HUB::Invoking GET method for control ID: e7a7c755-647e-4b69-8853-d66af047be4c
** Translator::get for control ID: e7a7c755-647e-4b69-8853-d66af047be4c
```
As shown in the log, the control ID of outlet 1’s translator is overwritten by that of outlet 2 even though I was invoking the method using the translator created for outlet 1.

m-chau:
Fixes is included in PR #152 .

danielrischar:
This should be fixed for Wink, SmartThings, Nest, and Insteon now.  Just need to get some more eyes on it.

PR #152, opened 2017-02-06
Fixes on SmartThings translators.

m-chau:
includes:
- an up-to-date SmartApp code
- fixed on how translators returns control IDs.
- removed static variables in SmartThings, Nest, Wink, and Insteon translators that were causing Issue #142 
- fixed issues #153 and #154.


org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -93,34 +117,27 @@ def installed() {
 def updated() {
 	log.debug "Updated with settings: ${settings}"
 	unsubscribe()
-	initialize()
```
arjun-msft: >initialize() [](start = 1, length = 12)

was this intentional?

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -129,290 +146,239 @@ def registerChangeHandler(myList) {
...
-            	break
-            case "water sensor":
-            	deviceType = "waterSensor"
-            	break   
-            default:
-            	break
-        }
-    }
-    return deviceType
+				if (caps.any{it.name.toLowerCase() == "power meter"}){
```
arjun-msft: I don't follow the logic here, how can it.name.toLowerCase() be "switch" and "power meter" at the same time?

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -93,34 +117,27 @@ def installed() {
...
 def registerSubscriptions() {
-	registerChangeHandler(switches)
-	registerChangeHandler(motionSensors)
-    registerChangeHandler(presenceSensors)
-    registerChangeHandler(contactSensors)
-    registerChangeHandler(waterSensors)
-    registerChangeHandler(locks)
-    registerChangeHandler(garageDoors)
-    registerChangeHandler(cameras)
+	registerChangeHandler(inputs)
```
arjun-msft: >inputs [](start = 23, length = 6)

where does this value come from? inputs

arjun-msft:
:clock1:

shalinijoshi19:
> definition(

Why do we have a .groovy file here again ? is this a test app? if so should it be under tests? 

---
Refers to: org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy:1 in 14fa260. [](commit_id = 14fa260ac6fb2b8872bb4059fde2fd4e7ab25ebc, deletion_comment = False)

arjun-msft:
> definition(

This is the SmartThings "smart app" which gets loaded in to their system which exposes methods that we can call from our translators.

---
In reply to: [278478972](https://github.com/openT2T/translators/pull/152#issuecomment-278478972) [](ancestors = 278478972)

---
Refers to: org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy:1 in 14fa260. [](commit_id = 14fa260ac6fb2b8872bb4059fde2fd4e7ab25ebc, deletion_comment = False)

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
shalinijoshi19: +1 -- unless in groovy, the 'it' implicit parameter gets re-scoped to the current value in the new loop?? either ways this is too confusing.. both logic (switch capablity doesnt mean switch device type; it could also mean 'light' depending on presence of another cap (switch level) in the returned cap list) as well as confusing as far as implementation (this use of default 'it' parameter is confusing) 

---
In reply to: [100167232](https://github.com/openT2T/translators/pull/152#discussion_r100167232) [](ancestors = 100167232)

shalinijoshi19:
:clock1:

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -129,290 +146,239 @@ def registerChangeHandler(myList) {
...
+                	return deviceType
+                }
+                if (caps.any{it.name.toLowerCase() == "switch level"}){
+                	deviceType = "light"
+                    return deviceType
+                }
+				break
+			case "contact sensor":
+				deviceType = "contactSensor"
+                return deviceType
```
arjun-msft: >deviceType [](start = 23, length = 10)

why do we need a return deviceType for each case? doesn't it get returned at the end?

if you need it inside the case, why not do return "contactSensor" instead of the two lines you have now?

shalinijoshi19:
> definition(

what system? who loads it? 

---
In reply to: [278481222](https://github.com/openT2T/translators/pull/152#issuecomment-278481222) [](ancestors = 278481222,278478972)

---
Refers to: org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy:1 in 14fa260. [](commit_id = 14fa260ac6fb2b8872bb4059fde2fd4e7ab25ebc, deletion_comment = False)

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
danielrischar: That's how Groovy works.  'it' is the implicit name of the closure parameter if none is provided.  http://groovy-lang.org/closures.html

Since it can also be specified explicitly, it would definitely help a lot with readability if if caps.each(...) used an explicit one line caps.each( capability => ... ).

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
m-chau: Yes. since we donnot want to reset variables like connectionId and endpointURL.

---
In reply to: [100166841](https://github.com/openT2T/translators/pull/152#discussion_r100166841) [](ancestors = 100166841)

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
m-chau: it's from vairable getInputs(). The usage is recommended by Samsung seems to be specific to SmartApp where it automatically maps "get" methods to a variable name. I.e. if you have a method called getInputs(), you can simplify calls to that by calling "inputs.xxxx"

---
In reply to: [100167546](https://github.com/openT2T/translators/pull/152#discussion_r100167546) [](ancestors = 100167546)

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
m-chau: Since we will a list of capability, I thought it would faster to return once we determined the device type instead of continue search through the capability list. 

If I just return a string with assigning it to a variable, I would actually get a null object back. I guess Groovy is a little weird this way.

---
In reply to: [100187469](https://github.com/openT2T/translators/pull/152#discussion_r100187469) [](ancestors = 100187469)

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
m-chau: Followed Daniel's advise in making the code more readable and added some comments as well.

---
In reply to: [100187044](https://github.com/openT2T/translators/pull/152#discussion_r100187044) [](ancestors = 100187044,100167232)

m-chau: Fixed according to suggestion.

---
In reply to: [100189818](https://github.com/openT2T/translators/pull/152#discussion_r100189818) [](ancestors = 100189818)

m-chau:
Added a fix to remove static variables in SmartThings translators class that were causing Issue #142

danielrischar:
:shipit:

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -129,290 +146,243 @@ def registerChangeHandler(myList) {
...
-            case "water sensor":
-            	deviceType = "waterSensor"
-            	break   
-            default:
-            	break
-        }
-    }
-    return deviceType
+                //If the device also contains "Power Meter" capability, identify it as a "switch" device.
+				if (capabilities.any{it.name.toLowerCase() == "power meter"}){
```
shalinijoshi19: >== [](start = 47, length = 2)

===

org.opent2t.sample.hub.superpopular/com.smartthings.hub/js/groovy/SmartApp.groovy
```diff
@@ -129,290 +146,243 @@ def registerChangeHandler(myList) {
...
+                	return deviceType
+                }
+                //If the device also contains "Switch Level" capability, identify it as a "light" device.
+                if (capabilities.any{it.name.toLowerCase() == "switch level"}){
```
shalinijoshi19: >== [](start = 59, length = 2)

===

PR #170, opened 2017-02-17
Ensuring required schema properties on translators

danielrischar:
Adds manufacturer name (mnmn), model (mnmo), entity name, spec version, and data model version if they are not provided by the device.


ChuckFerring:
:shipit:

PR #177, opened 2017-02-21
Updating package versions for Wink and SmartThings

danielrischar:
Ensuring versions in package.json are consistent with what is published on NPM.


ChuckFerring:
:shipit:

Issue #171, opened 2017-02-17
Insteon hub logs and ignores get_status failures


danielrischar:
If the get_status call in the Insteon hub translator fails, the error is logged to the console and ignored.  Line 100, and line 412.

These errors should be thrown so that the app can tell that the device is incommunicado (this happens if the device itself is turned off) rather than thinking that the communication is still okay.

m-chau:
Fix applied in https://github.com/openT2T/translators/pull/187.

PR #187, opened 2017-02-27
Insteon - Update how device type is determined.

m-chau:
- support more thermostat models.
- support dimming switch as lamps.
- passed unit tests.


org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
```diff
@@ -340,17 +342,29 @@ class Translator {
...
                     };
-                }
-                return undefined;
+                }// else a dimmer switch. 
+                
+                //Since dimmer switch schema is not ready, return light schema.
+                return {
+                    "schema": 'org.opent2t.sample.lamp.superpopular',
+                    "translator": "opent2t-translator-com-insteon-lightbulb"
+                };
```
danielrischar: We already merged a change that enabled the dimmer control as decCat 1, subCat E, so this is not needed and it casts too wide of a net, potentially trying to use a lightbulb translator for any device with a devCat of 1.

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
```diff
@@ -96,22 +96,24 @@ class Translator {
...
                                                 // Use get to translate the Insteon formatted device that we already got in the previous request.
                                                 // We already have this data, so no need to make an unnecesary request over the wire.
                                                 return OpenT2T.invokeMethodAsync(translator, opent2tInfo.schema, 'get', [expand, deviceData])
                                                     .then((platformResponse) => {
                                                         return platformResponse;
                                                     });
+                                            }).catch((err) => {
+                                                console.log('warning: OpenT2T.createTranslatorAsync error - ' + err);
+                                                return Promise.resolve(undefined);
                                             });
```
danielrischar: Don't catch this and let the rejected promise fall through to be caught by OpenT2T and converted into an OpenT2TError.

danielrischar: Oh, I see that it's a log with a promise.resolve, not a reject.  In that case leave it as is.

---
In reply to: [103301882](https://github.com/openT2T/translators/pull/187#discussion_r103301882) [](ancestors = 103301882)

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
```diff
@@ -96,22 +96,24 @@ class Translator {
...
                                         
+                                    }).catch((err) => { 
+                                        throw err;
                                     });
```
danielrischar: I'm not sure I follow the flow here.  If _getCommandResponse rejects, or throws, it will be caught here and re-thrown.  Does the uncaught exception get converted into a rejection?  We'd prefer to always explicitly reject if possible inside of asynchronous tasks.  Rejections will get converted into OpenT2TError classes by OpenT2T.

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
danielrischar: If we are comfortable choosing translators based on just devCat (1=lightbulb, 2=binaryswitch, 3=thermostat) then we should just make that change and get rid of the subCat white list completely.

---
In reply to: [103301675](https://github.com/openT2T/translators/pull/187#discussion_r103301675) [](ancestors = 103301675)

m-chau: I thought display dimmer switch a an light bulb is just an temporary solution before we get the dimmer switch schema? If we device to have a dimmer switch schema, then we would need to user subCat to differentiate.

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
```diff
@@ -398,18 +412,22 @@ class Translator {
...
+                                return this._getCommandResponse(commandId, --count);
+                            } else {
+                                this._throwError('Internal Error - Insteon command timeout.');
+                            }
+                            break;
+                        case 'succeeded':
+                            return Promise.resolve(response);
+                        case 'failed':
+                        default:
+                            this._throwError( 'Internal Error - Insteon command failed for deviceId ' + response.command.device_id + '.');
```
danielrischar: I believe that when a device is offline, that get_status will report "failed".  I see this happening now with some devices, and it causes getPlatforms to fail.

m-chau: Yes, that is why I only log the failure for _providerSchemaToPlatformSchema before. However, I do agree that we should let user know there is a communication problem. Should we have another resource to indicate if a device is online/offline?

---
In reply to: [103536052](https://github.com/openT2T/translators/pull/187#discussion_r103536052) [](ancestors = 103536052)

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
m-chau: I thought display dimmer switch a an light bulb is just an temporary solution before we get the dimmer switch schema? If we device to have a dimmer switch schema, then we would need to user subCat to differentiate.

---
In reply to: [103306611](https://github.com/openT2T/translators/pull/187#discussion_r103306611) [](ancestors = 103306611,103301675)

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
m-chau: Got it. I will change it to Promist.reject(err) instead.

---
In reply to: [103302937](https://github.com/openT2T/translators/pull/187#discussion_r103302937) [](ancestors = 103302937)

m-chau:
Added fix for https://github.com/openT2T/translators/issues/171 by adding a 'connectionStatus' resource.

org.opent2t.sample.binaryswitch.superpopular/com.insteon.binaryswitch/js/thingTranslator.js
```diff
@@ -106,6 +118,7 @@ function resourceSchemaToProviderSchema(resourceId, resourceSchema) {
         case 'n':
             result['DeviceName'] = resourceSchema.n;
             break;
+        case 'connectionStatus':
         default:
             // Error case
             throw new Error("Invalid resourceId");
```
shalinijoshi19: Youll likely need to merge tthis with Dan's changes once they go in today

shalinijoshi19:
:shipit:

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
avneet723: For now, let's remove the subcat. We want all dimmer switches to be casted as bulbs.

org.opent2t.sample.binaryswitch.superpopular/com.insteon.binaryswitch/js/thingTranslator.js
```diff
@@ -69,9 +69,20 @@ function providerSchemaToPlatformSchema(providerSchema, expand) {
         if: ['oic.if.a', 'oic.if.baseline']
     };
     
+    // Build the connectionStatus resource (read-only)
+    var connectionStatus = {
```
danielrischar: I like this solution, and it will work well for device graph updates.  Since this is additive, I see no issue with merging it, but we need to update the RAML (at least) and all of the other translators that will support this resource before it can be consumed.

danielrischar:
:shipit:

org.opent2t.sample.hub.superpopular/com.insteon.hub/js/thingTranslator.js
avneet723: Issue filed: Perform a subCat cleanup across all Insteon translators #194

PR #196, opened 2017-03-04
Throwing opent2constant string in wink

danielrischar:
Switches to throwing a message from OpenT2TConstants

### DIFF
--- a/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
+++ b/org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
@@ -77,7 +77,7 @@ class Translator {
                 var hashFromPayload = this._generateHmac(verification.key, payloadAsString);
 
                 if (hashFromWink !== hashFromPayload) {
-                    throw new OpenT2TError(401, "Notification signature doesn't match.  Expecting " + hashFromWink + " and calculated " + hashFromPayload);
+                    throw new OpenT2TError(401, OpenT2TConstants.HMacSignatureVerificationFailed);
                 }
             }
 


org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -77,7 +77,7 @@ class Translator {
                 var hashFromPayload = this._generateHmac(verification.key, payloadAsString);
 
                 if (hashFromWink !== hashFromPayload) {
-                    throw new OpenT2TError(401, "Notification signature doesn't match.  Expecting " + hashFromWink + " and calculated " + hashFromPayload);
+                    throw new OpenT2TError(401, OpenT2TConstants.HMacSignatureVerificationFailed);
```
shalinijoshi19: The specific was fine yes ? why change this instead of the test?

shalinijoshi19: What you can do here is to concatenate the constant with the "Expecting" part;
And the test in the cloud repo can then be updated with looking for (contains()) OpenT2TConstants.HmacSignatureVerificationFailed.

danielrischar: The hashes themselves don't provide any added value to the error massage as they are not usefulin debugging the issue.  The caller already has all of the info that they need to figure out the problem (their original key, and the payload).

shalinijoshi19: i see --- maybe log the error with teh specifics (though without logging hte input logging the hash seems useless also) for debugging help/telemetry) but we can tackle that later.

PR #199, opened 2017-03-07
Dynamic URL support for getUserPermissions

danielrischar:
Changes the SmartThings hub to use dynamic URL generation.  This frees up the caller from having to perform string replacement, and opens the door for dynamic URLs that come from other sources.

Nest also uses an oauth2 verification URI, but I've left it as is in order to allow both the old and new versions to co-exist in the CLI.


shalinijoshi19:
:shipit:

ChuckFerring:
:shipit:

arjun-msft:
:shipit:

PR #201, opened 2017-03-08
Add "all platform" unit tests

danielrischar:
Adds a new set of tests for all platform unit tests, and moves the ValidTranslator, GetPlatform, GetPlatformExpanded to this new group of tests.  Any tests that apply to all translators can now go here and will automatically be run in each set of unit tests.

Adds a ton of verification to GetPlatform to ensure that all translators return certain required properties (and found some translators that didn't).  This verification will be most useful when developing new translators as a way of ensuring that the schema output meets some minimal requirements.

The tests use snapshots to store the response to getPlatforms based on the testData.  If the translator or testData change, the unit test will fail unless the snapshot is also updated, ensuring that any changes to the translation are done intentionally.

Stubs out subscription test methods, which are skipped for now, they'll be expanded on next.


org.opent2t.sample.lamp.superpopular/com.smartthings.lightbulb/js/tests/unitTests.js
```diff
@@ -7,6 +7,7 @@ var MockHub = require('opent2t-device-smartthingshub/mockSmartthingsHub');
 var mockHub = new MockHub(deviceData);
 
 var settings = {
+    name: "com.smartthings.lightbulb",
```
ChuckFerring: >name: "com.smartthings.lightbulb", [](start = 4, length = 34)

Is this actually used?

danielrischar: Nope, left over from some other experiments.  I'll remove.

org.opent2t.sample.binaryswitch.superpopular/test/binarySwitchTests.js
```diff
@@ -1,19 +1,24 @@
...
     var deviceId = settings.deviceId;
+    settings.schemaName = SchemaName;
 
     function setData(t) {
         if(settings.setTestData) {
             settings.setTestData(t.title, t);
         }
     }
 
+    runAllPlatformTests(settings);
```
ChuckFerring: >runAllPlatformTests(settings); [](start = 4, length = 30)

Nice that you're able to add these here.

unittests/tests.js
```diff
@@ -0,0 +1,106 @@
...
+    var SchemaName = settings.schemaName;
+    
+    test.before(() => {
+        return settings.createTranslator().then(trans => {
+            translator = trans;
+            return OpenT2T.invokeMethodAsync(translator, SchemaName, 'get', []).then((response) => {
+                if(deviceId === undefined) {
+                    deviceId = response.entities[0].di;
+                }
+			});
```
ChuckFerring: Nit: change to spaces.

ChuckFerring:
:shipit:

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/thingTranslator.js
```diff
@@ -79,7 +79,7 @@ function providerSchemaToPlatformSchema(providerSchema, expand) {
         entities: [
             {
                 n: providerSchema['name'],
-                ivc: "core.1.1.0",
+                icv: "core.1.1.0",
```
shalinijoshi19: >icv [](start = 16, length = 3)

huh -- so how was this working to begin with ? Is this because we never tested this and are catching it now because of your changes or because of some recent prod change?

shalinijoshi19:
:shipit:

org.opent2t.sample.binaryswitch.superpopular/com.smartthings.binaryswitch/js/thingTranslator.js
danielrischar: We never tested it, and we don't have multiple core versions at the moment, so I doubt anybody looks at this value today.

org.opent2t.sample.binaryswitch.superpopular/com.insteon.binaryswitch/js/tests/__snapshots__/unitTests.js.snap
```diff
@@ -0,0 +1,3 @@
...
+
+exports[`GetPlatformExpanded 1`] = `"{"opent2t":{"schema":"org.opent2t.sample.binaryswitch.superpopular","translator":"opent2t-translator-com-insteon-binaryswitch","controlId":"Mock Device"},"pi":"44b0a5fb-f8a2-0160-0fae-34978b82a172","mnmn":"Insteon","mnmo":"Binary Switch (Generic)","n":"Mock Device","rt":["org.opent2t.sample.binaryswitch.superpopular"],"entities":[{"n":"Mock Device","icv":"core.1.1.0","dmv":"res.1.1.0","rt":["oic.d.smartplug"],"di":"f9604075-1a64-498b-ae9b-7436a63721ba","resources":[{"href":"/power","rt":["oic.r.switch.binary"],"if":["oic.if.a","oic.if.baseline"],"id":"power","value":false},{"href":"/connectionStatus","rt":["oic.r.mode"],"if":["oic.if.s","oic.if.baseline"],"id":"connectionStatus","supportedModes":["online","offline","hidden","deleted"],"modes":["offline"]}]}]}"`;
```
arjun-msft: did you mean to include .snap files?

danielrischar: Yes, these are snapshots of our Mock device state that Ava can generate.  If the translator or the test data change, and affect the output of getPlatforms then the snapshot comparison will fail unless it is also updated to reflect the new info.

As we make modifications to the translators, we want to be able to verify that the platform scheme output from get(...) doesn't accidentally change.

On failure, Ava will automatically log a diff against the snapshot, and explain what can be done to check in an updated version if the change was intentional.

PR #208, opened 2017-03-13
Correcting/updating references to onboarding

arjun-msft:


arjun-msft:
Not sure why the CIs are failing, nothing to do with my changes

danielrischar:
CI failure is caused by a bad SmartThings lightbulb snapshot as a result of previous merge ordering.  It's fixed in https://github.com/openT2T/translators/pull/209

danielrischar:
:shipit:

PR #209, opened 2017-03-13
Adding colour support to the Wink light bulb translator

danielrischar:
Adds colour support (OCF uses the British spelling) to the Wink bulb translator.


danielrischar:
This breaks the SmartThings color unit tests since it fixes the casing of rgbValue in lampTests.js.  I'll add those fixes here tout suite.

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -44,107 +51,476 @@ function findResource(schema, di, resourceId) {
...
 /**
- * Wink does not always populate every desired_state property, but last_reading doesn't necessarily
- * update as soon as we send our PUT request. Instead of relying just on one state or the other,
- * we use this StateReader class to read from desired_state if it is there, and fall back to last_reading
- * if it is not.
+ * Validates that an object (arguments[0]) contains properties
+ * (arguments[1...n]) and throws a descriptive OpenT2T error if the
+ * request is malformed.
+ * 
+ * TODO: This method should be moved to a shared location for all translators
```
arjun-msft: >This method should be moved to a shared location for all translators [](start = 9, length = 68)

Are we blocked on this at all?

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
```diff
@@ -44,107 +51,476 @@ function findResource(schema, di, resourceId) {
...
 
 /**
- * Helper method to convert a 0-100 scale to Wink's 0.0-1.0 brightness scale
+ * Create a colourRgb resource
+ * 
+ * TODO: This method should be moved to a shared location for all translators
  */
-function scaleTranslatorBrightnessToDeviceBrightness(dimmingValue) {
-    return dimmingValue / 100;
+function createColourRGBResource(expand, rgbArray, range = [0, 255]) {
```
arjun-msft: >range = [0, 255] [](start = 51, length = 16)

can range ever be anything else for rgb? not sure if this needs to be a parameter

org.opent2t.sample.hub.superpopular/com.wink.hub/test/mockWinkHub.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+var _ = require('lodash');
```
arjun-msft: why are we adding this dependency?

arjun-msft:
:clock1:

org.opent2t.sample.hub.superpopular/com.wink.hub/test/mockWinkHub.js
danielrischar: I wanted the deep equals functionality (line 16 below).  lodash is used by Ava for this purpose already, so it doesn't add any additional packaged to node_modules.

---
In reply to: [106019229](https://github.com/openT2T/translators/pull/209#discussion_r106019229) [](ancestors = 106019229)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
danielrischar: No, I simply wanted to tag which methods I'm going to be pulling out into a shared library at some point in the near future.

---
In reply to: [106017868](https://github.com/openT2T/translators/pull/209#discussion_r106017868) [](ancestors = 106017868)

org.opent2t.sample.lamp.superpopular/com.wink.lightbulb/js/thingTranslator.js
danielrischar: The colourRGB resource has a range property, though if it is omitted we should assume [0, 255].  In this case, Wink won't provide it as anything different, but this method is going to be moved to a shared location, and somebody may want to expose a different range depending on the provider.

---
In reply to: [106018057](https://github.com/openT2T/translators/pull/209#discussion_r106018057) [](ancestors = 106018057)

Issue #207, opened 2017-03-13
Wink sensor doesn't handle undefined timestamp values


arjun-msft:
There are times when undefined (we think) gets returned from Wink as a result the sensor errors out

arjun-msft:
Error seen:
RangeError: Invalid time value

PR #210, opened 2017-03-14
Fix for WINK Sensor translator - https://github.com/openT2T/translators/issues/207

shalinijoshi19:
(1) Updating the translator to only convert valid datetime values and return others (undefined/null/empty/non-number strings) as-is without conversion.
(2) Updating Unit Tests to run against an invalid data set for the _changedat_ property (added new data file for this)
(3) Fixing a couple of minor issues in the translator


.gitignore
```diff
@@ -12,4 +12,5 @@ testconfig.json
 /package.json
 
 // Private Tarballs
-*.tgz
\ No newline at end of file
+*.tgz
+*.snap
```
danielrischar: Snapshots are part of our unittesting.  If git shows that they are modified, then that means either the deviceData.json or the translator output changed.  If that was expected, then submit the new version.  It's a test to guarantee we don't accidentally add or remove anything from the translator output.

shalinijoshi19: I see -- I assumed these had something to do with the debugging I was doing and VS Code related files.. I will remove the .gitignore.

danielrischar:
:clock1:

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/tests/invalidDeviceData.json
```diff
@@ -0,0 +1,118 @@
...
+                "tamper_detected_true_updated_at": null,
+                "connection": true,
+                "connection_updated_at": 1488847336.5081685,
+                "agent_session_id": null,
+                "agent_session_id_updated_at": null,
+                "connection_changed_at": 1488846785.7296247,
+                "motion_changed_at": "undefined",
+                "battery_changed_at": "N/A",
+                "temperature_changed_at": 1488847336.5081685,
+                "tamper_detected_changed_at": 1488847025.6141558,
```
shalinijoshi19: Would love to trim it down to just invalid data for *_changed_at but i wasnt sure which ones were absolutely needed for the translator and which ones were extraneous here.

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/thingTranslator.js
```diff
@@ -84,7 +89,7 @@ function convertDeviceBatteryToTranslatorBattery(batteryValue) {
  * Returns a default value if the specified property is null, undefined, or an empty string
  */
 function defaultValueIfEmpty(property, defaultValue) {
-    if (property === undefined || property === null || property === "") {
+    if (!property) {
```
begillil: This check (`if (!property)`) might use the default value in unexpected situations. If the property you're reading is a number, we would use the default if you're reading the value '0', or if the property you're reading is a Boolean value, we would use the default if you're reading the value 'false'.

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

shalinijoshi19: I see -- ok will revert here. 

---
In reply to: [106045877](https://github.com/openT2T/translators/pull/210#discussion_r106045877) [](ancestors = 106045877)

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/tests/invalidDeviceData.json
```diff
@@ -0,0 +1,118 @@
...
+                "tamper_detected_true_changed_at": 1488846998.8763678,
+                "motion_true_changed_at": "undefined",
```
begillil: I don't think Wink would return the string 'undefined' in its payload. My understanding is that there might be cases where the _changed_at field is excluded from the payload entirely, which would ultimately lead to us passing the JavaScript value undefined (rather than the string "undefined") to the convertDeviceDateToTranslatorDate function.

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/thingTranslator.js
```diff
@@ -70,10 +70,15 @@ class StateReader {
...
+    // Only convert if input is valid numeric value
+    if (unixTimestamp && !isNaN(unixTimestamp)) { 
+        // Date takes a number of milliseconds, so convert seconds to milliseconds
+        var datetime = new Date(unixTimestamp * 1000);
+        return datetime.toISOString();
+    }
 
-    return datetime.toISOString();
+    // return all other invalid input as-is since we can't convert to DateTime
+    return unixTimestamp;
```
begillil: >// return all other invalid input as-is since we can't convert to DateTime 

The lastchanged resource is defined as an opent2t.r.timestamp, which specifies that the string should be in the ISO8601 format.

`"description": "Using ISO 8601 datetime format (e.g: 2007-04-05T14:30Z, 2007-04-05T14:30+09:00)"`

It would be good to understand why the _changed_at property isn't always showing up in the Wink payload. If it happens for devices that haven't reported a value yet, for example, it might be a good idea to fall back to the _updated_at property instead of just returning 'undefined' here.

See the Wink documentation for the difference between the two:

http://docs.winkapiv2.apiary.io/reference/device/desired-state-and-last-reading

shalinijoshi19: WINK returns "undefined" - a string value - in some cases only not always - while its great to understand why WINK doesnt adhere to the standard i think our own code should also be robust here and handle at least the values that we are observing are being sent back and not blowing up. Here the issue is the Date constructor has been working for null returning an epoch value date; so us returning what the 3rd party provider is returning (whether they are buggy ornot ) seems to be least controversial and cooked up to me at hte moment. 

---
In reply to: [106047060](https://github.com/openT2T/translators/pull/210#discussion_r106047060) [](ancestors = 106047060)

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/tests/invalidDeviceData.json
shalinijoshi19: That's what Arjun confirmed was what being returned - this was causing the Date constructor to blow up. Eitherways this covers any possible un-translateable value that might be returned and which we dont know how to convert. 

---
In reply to: [106046809](https://github.com/openT2T/translators/pull/210#discussion_r106046809) [](ancestors = 106046809)

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/thingTranslator.js
begillil: Well, the standard is in our schema, not theirs. We should be responsible for ensuring we conform to our schema.

I'm suggesting that we initially try to use the _changed_at property (e.g. motion_changed_at) since that is the time that is reported by the physical device itself, and if Wink doesn't provide that for whatever reason (which Arjun saw in his test), we can fall back to the _updated_at property, which is set by Wink, rather than reported by the device.

If there are cases where neither of those properties exist, I would suggest we fall back to a valid ISO8601 datetime (e.g. the epoch time, or something similar), rather than returning something that doesn't conform to our schema.

shalinijoshi19: I see - thanks Ben esp because i am ramping up on this sensor type -- but our schema is likely derived off the WINK schema yes? i see your point however looks like updated_at and changed_at might not always be the same - and if WINK distinguishes shouldnt we too? Isnt it going to get very confusing if we start to cook up values that might not jive with what the user might see if they were to use a real WINK app?  But maybe i dont quite get the scenario yet  and what we are aiming for with this translation in the first place. If we (in opent2t) are flexible about conflating updatedAt/ChangedAt then sure i like the idea of falling back to udpated, but wouldnt that mean another query/lookup to the server for the property ? I havent yet investigated if we query for the entire object in one go or if use targeted queries;   Also I'd much rather go for simpler logic here but then again i am not sure what our goal is. I'll let the others esp Arjun weigh in also; I dont mind the default epoch time for the default case but it again might get confusing and misleading. 

I think this needs a broader discussion here -- this change at least unblocks hte package publish; What i think the next step could be to also add a telemetry event here to try to see what are the different cases where we fail to convert and re-evaluate our behavior/schema ?

---
In reply to: [106049051](https://github.com/openT2T/translators/pull/210#discussion_r106049051) [](ancestors = 106049051)

begillil: >but our schema is likely derived off the WINK schema yes?

The same schema is used to define sensors coming from all providers, not just Wink.

>i see your point however looks like updated_at and changed_at might not always be the same - and if WINK distinguishes shouldnt we too?

The difference between the two is that one is reported by the physical device, and the other is when Wink sees the value change (which, I agree, could be different).

This is why I wanted to understand better the cases where Wink doesn't provide the _changed_at value. Their documentation says that this value is reported by the physical device. This suggests to me that device Arjun saw this with may not have reported a value yet. If that is by design for Wink, then we need a sane way to handle that case. 

_updated_at seemed like a good proxy for the _changed_at value, since they should be similar. If the scenario I outlined above is the only time this can happen, another alternative might be to use the device's [created_at](http://docs.winkapiv2.apiary.io/reference/device) timestamp, since, essentially the last time the value changed would be when the device was created in that scenario.

I don't think we should get in the habit of just returning values directly from the third parties in cases like this, since (unless we get lucky) what they're giving us will not match the format that our schema specifies. Our clients expect the data coming back from the translators to match the schema we say we translate to (otherwise what's the point of the schema?).

> but wouldnt that mean another query/lookup to the server for the property

No, the translator expects the entire Wink device payload. See [devicedata.json](https://github.com/openT2T/translators/blob/dev/org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/tests/devicedata.json) for an example of a Wink device payload (id fields were changed, and fields were added to show examples of every sensor value defined in [Wink's API](http://docs.winkapiv2.apiary.io/reference/device/sensor), but should otherwise be about what you should get. Note that a typical sensor would only have properties for the types of sensors included in that device, e.g. the GoControl motion sensor has the 'motion' field and the 'temperature' field, along with the motion_updated_at, etc fields. It wouldn't have the others).

org.opent2t.sample.multisensor.superpopular/test/multisensorTests.js
```diff
@@ -284,6 +284,17 @@ function runMultisensorTests(settings) {
             });
     });
 
+        test.serial('GetBrightnessChange', t => {
```
begillil: Isn't there already a test for brightnesschange? Line 101?

shalinijoshi19: Doh! thanks missed seeing it up there. 

---
In reply to: [106059745](https://github.com/openT2T/translators/pull/210#discussion_r106059745) [](ancestors = 106059745)

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/thingTranslator.js
arjun-msft: For this case this method should always return a DateTime or null/undefined.  It *should not* return a string.  So my opinion is we have two options, in the case where we can't create a DateTime based on *_changed_at:

1. we return null/undefined
2. we return *_updated_at

I'm ok with either, 2 *may* give us a more consistent return but also *may* have some delayed value but given this case it may be the only value we'll get.  I also recommend moving forward with option 2 given Wink's docs.

---
In reply to: [106057659](https://github.com/openT2T/translators/pull/210#discussion_r106057659) [](ancestors = 106057659)

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/tests/invalidDeviceData.json
arjun-msft: I don't know exactly what was returned, I know the error showed up. My guess is it was missing and resulted in javascript undefined.

---
In reply to: [106048032](https://github.com/openT2T/translators/pull/210#discussion_r106048032) [](ancestors = 106048032,106046809)

arjun-msft:
:clock1:

org.opent2t.sample.multisensor.superpopular/com.wink.sensorpod/js/thingTranslator.js
shalinijoshi19: Thanks guys - so I've gone with : 
(1) return _changed_at where available
(2) if not a valid number or undefined, we go with _updated_at
(3) Else return undefined.

PR #212, opened 2017-03-14
Fixing Wink colourMode names

danielrischar:
Wink bulb translator was returning colourMode.supportedModes with Wink naming rather than resource naming.


shalinijoshi19:
:shipit:

Issue #232, opened 2017-04-13
Some switches in SmartThings are coming back with brightness scale in 0-255


arjun-msft:
The AEO switch on our account is showing up as having a brightness of 255.

PR #235, opened 2017-04-19
Add a check for max dimming value ST lights.

m-chau:
A fix for #232 :  The root cause of the issue seems to be between AEOTec and SmartThings. When we turn the AEOTec off and on, the switch level sometimes (over 50% of time) showed up as 255% then to 99% (which is the max level in AEOTec switch) after a few seconds. Therefore, if we query for device status during this time frame, we will get dimming level of 255%.  This PR includes a workaround of this bug by placing a cap on the maximum brightness as light bulb can have.  According to SmartThings documentation, the switch level of a SmartThings device is usually from 0 to 100 in percent.

### DIFF
--- a/org.opent2t.sample.lamp.superpopular/com.smartthings.lightbulb/js/thingTranslator.js
+++ b/org.opent2t.sample.lamp.superpopular/com.smartthings.lightbulb/js/thingTranslator.js
@@ -207,6 +207,7 @@ function providerSchemaToPlatformSchema(providerSchema, expand) {
 
         dim.id = 'dim';
         dim.dimmingSetting = providerSchema['attributes'].level;
+        if (dim.dimmingSetting > 100 ) dim.dimmingSetting = 100;
         dim.range = [0, 100];
         
         if (supportColour) {


org.opent2t.sample.lamp.superpopular/com.smartthings.lightbulb/js/thingTranslator.js
```diff
@@ -207,6 +207,7 @@ function providerSchemaToPlatformSchema(providerSchema, expand) {
 
         dim.id = 'dim';
         dim.dimmingSetting = providerSchema['attributes'].level;
+        if (dim.dimmingSetting > 100 ) dim.dimmingSetting = 100;
```
arjun-msft: shouldn't we do a conversion here?  meaning what if its 125, wouldn't we want to set it to 50 in that case?  That requires us to guess, do we have data to back up that guess?

arjun-msft:
:clock1:

org.opent2t.sample.lamp.superpopular/com.smartthings.lightbulb/js/thingTranslator.js
m-chau: When I tested AEOTec switch through SmartThings app and checked the result in the ST portal, I did not see any level changes in the log for AOETec switch(the ST app's UI gives the illusion that level is adjusted , but in the backend it was not). The only time I see the dimming level change is between 99% and 255%, and when that happens, the ST always indicate the switch level is at maximum. Therefore, I am just capping the level occuring to the ST documentation (please see the description for this PR).

I am not sure it that weird behavior is due to fact that the switch is not outputting electricity. Maybe you can try to test it with the real hardware. You may check how AEOTec switch really behaves by selecting "AEO switch back wall" in the "live logging" tab of the ST portal.

---
In reply to: [112323616](https://github.com/openT2T/translators/pull/235#discussion_r112323616) [](ancestors = 112323616)

arjun-msft: Thanks, will try it out with hardware.

---
In reply to: [112694398](https://github.com/openT2T/translators/pull/235#discussion_r112694398) [](ancestors = 112694398,112323616)

PR #239, opened 2017-05-04
Change  "_subscribe" to "postSubscribe" to make Wink translator.

m-chau:


org.opent2t.sample.binaryswitch.superpopular/com.insteon.binaryswitch/js/thingTranslator.js
```diff
@@ -185,7 +185,7 @@ class Translator {
     }
 
     postSubscribe(subscriptionInfo) {
-        return this.insteonHub._subscribe(subscriptionInfo);
```
arjun-msft: insteon doesn't have subscribe, what happens in this scenario?

m-chau: I added the subscribe functions in the hub translator. It should return an error like Nest if it's called.

---
In reply to: [114919066](https://github.com/openT2T/translators/pull/239#discussion_r114919066) [](ancestors = 114919066)

PR #254, opened 2017-08-16
Adding support for deauthorizeToken

EllenPorter:


org.opent2t.sample.hub.superpopular/com.nest.hub/js/thingTranslator.js
```diff
@@ -56,6 +56,24 @@ class Translator {
...
+     */
+    deauthorizeToken(authInfo) {
+        
+        var options = {
+            url: 'https://api.home.nest.com/oauth2/access_tokens/' + this._authTokens['access'].token,
+            method: 'DELETE'
+        };
+
+        return request(options).then((body) => {
+            return true;
```
kurstory: If this request fails, can we have it return false?

EllenPorter: Yup, good idea, fixed.

org.opent2t.sample.hub.superpopular/com.wink.hub/js/thingTranslator.js
```diff
@@ -178,6 +178,17 @@ class Translator {
     }
 
     /**
+     * Adding for consistency across hubs
+     */
+    deauthorizeToken(authInfo) {
+        if (!authInfo) {
```
kurstory: Why is this distinction only in Wink and not the other providers?

EllenPorter: The Travis check marked the unused variable as an error - looking now to see why it was an error not warning like the other providers. If we implement deauthorize in the other providers we might need the authInfo, so I'd like to keep the variable being passed in.

PR #261, opened 2017-10-09
Hot fix for autoheat, autocool support in smartthings

EllenPorter:
Logs show users have thermostats with an 'autocool' or 'autoheat' mode, we currently set the wrong attribute for thermostats in this mode, they need to be treated like 'auto' mode.

### DIFF
--- a/org.opent2t.sample.thermostat.superpopular/com.smartthings.thermostat/js/thingTranslator.js
+++ b/org.opent2t.sample.thermostat.superpopular/com.smartthings.thermostat/js/thingTranslator.js
@@ -448,7 +448,6 @@ class Translator {
                             Object.assign(response, putPayload.response);
                             var schema = providerSchemaToPlatformSchema(response, true);
                             // from logs: autocool does not allow changing thermostat setpoint - we'll do target high/low for this case
-                            console.log(response.thermostatMode);
                             if (response.thermostatMode === 'auto' ||
                                 response.thermostatMode === 'autoheat' ||
                                 response.thermostatMode === 'autocool') {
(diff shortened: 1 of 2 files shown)


org.opent2t.sample.thermostat.superpopular/com.smartthings.thermostat/js/thingTranslator.js
```diff
@@ -447,7 +447,11 @@ class Translator {
                         return this.smartThingsHub.putDeviceDetailsAsync(this.endpointUri, this.controlId, putPayload.command).then((response) => {
                             Object.assign(response, putPayload.response);
                             var schema = providerSchemaToPlatformSchema(response, true);
-                            if (response.thermostatMode === 'auto') {
+                            // from logs: autocool does not allow changing thermostat setpoint - we'll do target high/low for this case
```
kurstory: Are there any issues setting high in autoHeat mode or low in autoCool mode?

kurstory:
Are there any issues setting high in autoHeat mode or low in autoCool mode?

EllenPorter:
There may be issues setting high or low but we don't have a thermostat to test with.  This will get us past the error we are seeing while trying to set the targetSetpoint but we may need a second hotfix if we now see errors setting high or low.  I am hoping this mode looks like 'auto' mode.

kurstory:
I'm also maybe a bit confused how this actually works?  Looking at postDeviceResource, it patches targetTemperature regardless, and the if statement just filters the response back from OpenT2T?
I think you also have to update ~line 601 to actually look for these new modes as well, otherwise I don't think the POST to SmartThings changes at all.

